PR #1, opened 2015-10-27
everything

davidchambers:
Closes plaid/sanctuary#75

I'd love feedback from @boxofrox and @davidchase, particularly on the README.

The module exports `is` in addition to the type variables `a` through `z` and `def` itself, but I decided not to document `is`. The only reason to export it is so Sanctuary can define [`S.is`][1] without duplicating code which *must* be defined in this module.

[1]: https://github.com/plaid/sanctuary#is


davidchase:
@davidchambers just from reading the README :heart: `Custom pseudotypes` & `Type variables` 
now its time to see how it works in the actual code :+1:

README.md
```diff
@@ -0,0 +1,167 @@
...
+evaluate to `'abcdef'`. It's nonsensical, though, to concatenate values of
+different types, so both arguments must be of the same type:
+
+```javascript
+const a = def.types.a;
+
+//    concat :: Semigroup a => a -> a -> a
+const concat = def('concat', [a, a], (x, y) => x.concat(y));
+```
+
```
soroushjp: This is fantastic. Overall, README is very clearly written and I didn't have any stumbling blocks going thru it with little to no background.

README.md
```diff
@@ -0,0 +1,167 @@
...
+`def` has no notion of type classes. The `concat` function defined above does
+not ensure that its arguments are of a type which provides a `concat` method.
```
boxofrox: Everything made sense up to this point.  I'm not certain if you're telling us how to use type variables in this example, or that we can't use type variables because *"... [the] above does not ensure ..."*.

I now understand that `def` cannot check for a `concat` method in `concat`'s  arguments, but the question, "so what does it do?" nags at me.  I want to say `def` is able to validate that both arguments are the same, but it's not clear to me how to polymorphically check for type similarities in javascript.

Of course, I could read the code at this point.  If you'd rather spare readers from that requirement, I'm happy with you telling me, *"However, `def` does ensure the types of both arguments to `concat` are the same"*, and I'll accept your sage wisdom that it just works that way.

davidchambers: Great point, @boxofrox. I agree that we should explain how we determine whether two values are of the same type.

Also, it might be clearer to give this paragraph its own subheading, __Type classes__, to avoid casting doubt on whether the `concat` definition above is valid use of `def`. So we would first introduce type variables, then ask the question of whether it's possible to constrain `a` by requiring it to implement [Semigroup][1], for example. The answer, currently, is no.


[1]: https://github.com/fantasyland/fantasy-land#semigroup

README.md
davidchambers: Thanks for the feedback, @soroushjp. :)

davidchambers:
> now its time to see how it works in the actual code

[Sanctuary's source][1] provides real-world use of `def`. Also, https://github.com/plaid/sanctuary/compare/master...dc-def shows how Sanctuary's internals are likely to change when we add sanctuary-def as a dependency. Being able to define pseudotypes in a self-contained manner is very nice. For example:

```javascript
//  RegexFlags :: { name :: String, test :: a -> Boolean }
var RegexFlags = {
  name: 'RegexFlags',
  test: function(x) {
    return R.type(x) === 'String' &&
           R.test(/^g?i?m?$/, x);
  }
};
```


[1]: https://github.com/plaid/sanctuary/blob/master/index.js

README.md
boxofrox: > ...it might be clearer to give this paragraph its own subheading, Type classes, to avoid casting doubt on whether the concat definition above is valid use of def. So we would first introduce type variables, then ask the question...

:+1:

After reading the entire readme, I figured `Semigroup` could be implemented as a custom pseudotype if the only requirement was to check for a `concat` method.

davidchase:
> Sanctuary's source provides real-world use of def. Also, plaid/sanctuary@master...dc-def shows how 

I like the way you updated it with `sanctuary-def` easier to read and reason about IMHO

README.md
davidchambers: > After reading the entire readme, I figured `Semigroup` could be implemented as a custom pseudotype if the only requirement was to check for a `concat` method.

It could be, but we'd lose the requirement that both arguments be of the same type. See #3 for a possible solution.

davidchambers:
I've updated this pull request. Type classes are now supported! For example:

```javascript
//    Semigroup :: { name :: String, test :: a -> Boolean }
const Semigroup = {
  name: 'Semigroup',
  test: x => x != null && typeof x.concat === 'function',
};

//    concat :: Semigroup a => a -> a -> a
const concat = def('concat', {a: Semigroup}, [a, a], (x, y) => x.concat(y));
```

```javascript
concat([1, 2, 3], [4, 5, 6]);
// => [1, 2, 3, 4, 5, 6]

concat('abc', 'def');
// => 'abcdef'

concat(42);
// ! TypeError: ‘concat’ requires a Semigroup as its first argument; received 42

concat([], '')
// ! TypeError: ‘concat’ requires its first and second arguments to be of the same type; [] and "" are not
```

boxofrox:
Very amazing! :astonished:

davidchambers:
Superseded by #4

PR #4, opened 2015-11-16
everything

### LINKED ISSUE #3: type classes
#1 currently provides `forall a`:

```javascript
//    I :: a -> a
const I = def('I', [a], x => x);
```

But we have no way to constrain a type variable. Consider `concat`:

```javascript
//    concat :: a -> a -> a
const concat = def('concat', [a, a], (x, y) => x.concat(y));
```

We'd like to require that `a` have a [semigroup][1]:

```haskell
concat :: Semigroup a => a -> a -> a
```

We could define a Semigroup pseudotype, to enable:

```javascript
//    concat :: Semigroup -> Semigroup -> Semigroup
const concat = def('concat', [Semigroup, Semigroup], (x, y) => x.concat(y));
```

This loses the constraint that both arguments be of the same type, so nonsensical operations such as `concat([], '')` would be permitted.

Perhaps type constraints could be expressed via a dictionary mapping identifier to type class:

```javascript
const Semigroup = {
  name: 'Semigroup',
  test: x => x != null && typeof x.concat === 'function',
};

//    concat :: Semigroup a => a -> a -> a
const concat = def('concat', {a: Semigroup}, [a, a], (x, y) => x.concat(y));
```

[1]: https://github.com/fantasyland/fantasy-land#semigroup

### LINKED ISSUE #5: check return types
Currently the two open pull requests allow a function's *input* types to be specified, but not its *output* type:

```javascript
//    add :: Number -> Number -> Number
const add = def('add', {}, [$.Number, $.Number], (x, y) => 'XXX');
```

We could require the output type as well:

```javascript
//    add :: Number -> Number -> Number
const add = def('add', {}, [$.Number, $.Number, $.Number], (x, y) => 'XXX');
```

This would ensure that each function either adheres to its contract or fails noisily. Contract breaches would result in helpful exceptions:

```javascript
add(2, 2);
// ! TypeError: ‘add’ is expected to return a value of type Number; returned "XXX"
```

davidchambers:
Closes plaid/sanctuary#75; closes #3; closes #5

#1 was a good first effort, but when I updated Sanctuary itself to use it there were a few constraints I could not express. The type of `Just#concat`, for example, is `Semigroup a => Maybe a ~> Maybe a -> Maybe a`, yet `[Maybe]`—the description of that method's argument types—does not even *mention* `a`. This pull request makes it possible to describe arbitrarily nested types. For example:

```javascript
//    fromMaybe :: a -> Maybe a -> a
const fromMaybe =
def('fromMaybe', {}, [a, Maybe(a), a], (x, m) => m.isJust ? m.value : x);

fromMaybe(0, Just(42));
// => 42

fromMaybe(0, Nothing);
// => 0

fromMaybe(0, Just('XXX'));
// ! TypeError: ‘fromMaybe’ expected a value of type (Maybe Number) as its second argument; received Just("XXX")
```

The pull request also introduces the ability to define record types. For example:

```javascript
//    Point :: Type

//    dist :: Point -> Point -> Number
const dist = def('dist', {}, [Point, Point, $.Number], (p, q) =>
  Math.sqrt(Math.pow(p.x - q.x, 2) + Math.pow(p.y - q.y, 2))
);

dist({x: 0, y: 0}, {x: 3, y: 4});
// => 5

dist({x: 0, y: 0}, {x: 3, y: 4, color: 'red'});
// => 5

dist({x: '0', y: '0'}, {x: '3', y: '4'});
// ! TypeError: ‘dist’ expected a value of type { x :: Number, y :: Number } as its first argument; received {"x": "0", "y": "0"}
```

When reviewing #1 @Bpless suggested that special treatment of a value's `'@@type'` property is too opinionated for a package which needn't be used in conjunction with Sanctuary. This pull request does not treat this property specially (though it is used by data constructors in the test suite).

I'd :heart: feedback on this pull request. Specifically, I'd appreciate answers to the following questions:

  - Does the README read well? Does it cover all the concepts and functions?
  - Do the error messages read well?
  - Is `???` the optimal string representation of `UnknownType`?

/cc @albrow, @benperez, @boxofrox, @buzzdecafe, @CrisFeo, @CrossEye, @davidchase, @joneshf, @raine, @scott-christopher, @soroushjp, @svozza

_I won't make a habit of mentioning a dozen people in a pull request description! If I mentioned you above I did so because I value your opinions and I believe you will find this pull request interesting._


README.md
```diff
@@ -0,0 +1,492 @@
...
+
+Applying `add` to greater than two arguments results in an exception being
+thrown:
+
+```javascript
+add(2, 2, 2);
+// ! TypeError: ‘add’ requires two arguments; received three arguments
+```
+
+Applying `add` to fewer than three arguments results in a function awaiting
```
joneshf: Is this supposed to be "two" instead of "three"?

README.md
```diff
@@ -0,0 +1,492 @@
...
+the remaining arguments:
+
+```javascript
+add(2)(2);
```
joneshf: This example doesn't seem to be for the description above.

README.md
```diff
@@ -0,0 +1,492 @@
...
+  if (y === 0) throw new Error('Cannot divide by zero');
+  return x % y;
+});
+
+rem(42, 5);
+// => 2
+
+rem(42, 0.5);
+// ! TypeError: ‘rem’ expected a value of type Integer as its second argument; received 0.5
+```
```
joneshf: This example is great, you could go on to show how to use `NullaryType` to define a non-zero integer, and take that as the second argument. Thus you needn't check for it in the function!

README.md
```diff
@@ -0,0 +1,492 @@
...
+
+//    Suit :: Type
+const Suit = $.NullaryType(
+  'sanctuary-def/Suit',
+  x => typeof x === 'string' && /^[\u2660\u2663\u2665\u2666]$/.test(x),
+  '\u2660'
+);
+
+//    Card :: Pair Rank Suit
+const Card = $Pair(Rank, Suit);
```
joneshf: Extra `$`?

README.md
davidchambers: Yes, it is. Thanks!

:zap:

README.md
davidchambers: Quite right! Fixed.

:zap:

README.md
davidchambers: Terrific idea!

README.md
davidchambers: In this case `$Pair` is the type constructor; `Pair` is the data constructor. Perhaps it would be clearer to use `Card :: Type` as the type signature. What do you think?

joneshf:
This is amazing! I can't wait to see how far you take this.

* The readme reads nicely. I am a bit confused on some of the notation though, like what `*` and `_` are in type signatures.
* The error messages read nicely as well. I could see an improvement being taking into account both sides of the unification. It looks as though the checker currently chooses one type and attempts to unify another. But the problem might be that the first type the checker sees was input incorrectly by the user, instead of the second type being incorrect. Does that make sense?
* `???` seems like a fine choice to me. What's important is to let people know what it means. I didn't see it mentioned in the readme though.

README.md
davidchambers: Updated!

davidchambers:
> This is amazing! I can't wait to see how far you take this.

Thank you, Hardy. :)

> The readme reads nicely. I am a bit confused on some of the notation though, like what `*` and `_` are in type signatures.

I'm pleased you called me out on this. They indicate that my thoughts are unclear, or that certain concepts don't map nicely to JavaScript, or both. Perhaps you'll be able to suggest improvements once I explain my half-baked thoughts. :)

> ```haskell
> UnaryType :: String -> (* -> Boolean) -> (-> t _) -> (t a -> [a]) -> Type -> Type
> ```

`* -> Boolean` is meant to represent a predicate that operates on a value of any type. We can't use `a -> Boolean` since `a` is used elsewhere. Would `x -> Boolean` be better, do you think?

`-> t _` is meant to indicate a thunk which returns a value of unary type `t`. Take `Array` as a concrete example. The thunk can be defined as `() => []`. I had been thinking that it would be misleading to write `-> t a`, as this would suggest `() => [0]` as the thunk for `[Number]` and `() => ['']` for `[String]`, preventing the same thunk from being used for every specialization of `(Array ???)`. Now that I think about it, though, `() => []` has type `forall a. -> (Array a)`, so it's entirely reasonable to write `-> t a` rather than `-> t _`.

> The error messages read nicely as well. I could see an improvement being taking into account both sides of the unification. It looks as though the checker currently chooses one type and attempts to unify another. But the problem might be that the first type the checker sees was input incorrectly by the user, instead of the second type being incorrect. Does that make sense?

Could you provide an example? I'm not sure I follow.

> `???` seems like a fine choice to me. What's important is to let people know what it means. I didn't see it mentioned in the readme though.

Good point! I'll add a brief section to introduce `UnknownType`.

README.md
```diff
@@ -0,0 +1,495 @@
+# sanctuary-def
+
+sanctuary-def facilitates the definition of curried JavaScript functions which
+are explicit about the number of arguments to which they may be applied and the
+types of those arguments.
+
+```javascript
+const $ = require('sanctuary-def');
+
+const def = $.create($.env);
```
svozza: It's a bit confusing to have this $.env example here but the description of what it actually means right down the end of the the README.

scott-christopher:
This is great work, @davidchambers! I'm yet to review the implementation, but I'll throw a couple of general thoughts in here for now.

Have you considered capturing the return type of functions declared with `def`? Aside from the extra post-condition invariant, it should also enable the type signatures of those functions to be programmatically inferred for documentation, among other things.

Though I'm not yet entirely sure of the purpose of the arbitrary thunk function in the various type function arguments, if there is a benefit in the arbitrary values being random you could consider requiring them to be unary functions instead that accept a randomly generated number to derive the arbitrary value.

I'd also be interested to see how much of this could be erased during an optimisation step at build time for minified/production builds. One approach could just be to provide an alternative implementation of `def` which simply returns a curried version of the function.

Beyond that, I'm keen to have a play with it. :)

README.md
davidchambers: I agree that this is unfortunate. I like to begin a README with a straightforward example, and I don't see a straightforward way to convey why it's necessary to create an "environment". Perhaps we should write this:

```javascript
const def = $.create($.env);  // Explained in "Environments" section
```

I'm open to suggestions. :)

davidchambers:
> Have you considered capturing the return type of functions declared with `def`?

The idea occurred to me just a few hours ago. I opened #5 to track it, though I intend to experiment with incorporating return-type declarations into this branch.

> Aside from the extra post-condition invariant, it should also enable the type signatures of those functions to be programmatically inferred for documentation, among other things.

Great point! This could be very useful for keeping documentation up to date. :)

> I'm not yet entirely sure of the purpose of the arbitrary thunk function in the various type function arguments

The thunk provided access to an "exemplar", a value of the type which could be poked to see whether the type provides certain methods. Wrapping the value in a thunk proved convenient as one may wish to define types before defining data constructors.

The good news is that in the past few hours I figured out how to thread the *actual* values through the program, so it's no longer necessary to ask a type to provide an exemplar. No more thunks!

> I'd also be interested to see how much of this could be erased during an optimisation step at build time for minified/production builds. One approach could just be to provide an alternative implementation of `def` which simply returns a curried version of the function.

This is certainly something we should discuss. Someone suggested a `process.env.NODE_ENV === 'production'` check, which apparently works in the browser as well with certain build processes. Separate JavaScript files would be cleaner, though this would add complexity to the sanctuary-def project itself. This feature needn't be included in v0.1.0, of course. :)

> Beyond that, I'm keen to have a play with it. :)

Me too!

davidchambers:
In addition to removing the thunks, I pushed a change which means `[0, 0.5, 1] :: $.Array(Integer)` no longer type checks. Previously only the first element of the array was considered.

davidchambers:
I've pushed another change: return values are now type-checked!

```javascript
//    add :: Number -> Number -> Number
const add = def('add', {}, [$.Number, $.Number, $.Number], () => 'XXX');

add(2, 2);
// ! TypeError: ‘add’ is expected to return a value of type Number; returned "XXX"
```

I'm now happy with the functionality and interface provided by this pull request. The implementation can surely be improved, and the README still needs some attention. We're not too far away, though!

CrossEye:
So this is what you've been doing!  Very, very exciting!

README.md
```diff
@@ -0,0 +1,486 @@
...
+```haskell
+BinaryType :: String -> (* -> Boolean) -> (t a b -> [a]) -> (t a b -> [b]) -> Type -> Type -> Type
+```
+
+For example:
+
+```javascript
+//    $Pair :: Type -> Type -> Type
+const $Pair = $.BinaryType(
+  'sanctuary-def/Pair',
```
CrossEye: I'm just starting my way through this, but at first glance, I would expect that this section would be better with `'my-app/Pair'`, `'my-app/Rank'`, etc.  The namespace for these definitions are arbitrary but they probably **shouldn't** be in the `sanctuary-def` one, correct?

README.md
```diff
@@ -0,0 +1,486 @@
...
+_concat([1, 2], [3, 4]);
+// => [1, 2, 3, 4]
+
+_concat([1, 2], 'buzz');
+// ! TypeError: ‘_concat’ expected a value of type (Array Number) as its second argument; received "buzz"
+```
+
+The type of `_concat` is misleading: it suggests that it can operate on any
+two values of *any* one type. In fact there's an implicit constraint, since
+the type must support concatenation (in [Fantasy Land][2] terms, the type
```
CrossEye: Since `Semigroup` is already a link to the appropriate part of the FantasyLand specification, might it not be better to say

>    (in [mathematical][2] terms, the type must have a [semigroup][3])
>      [2]: https://en.wikipedia.org/wiki/Semigroup
>      [3]: https://github.com/fantasyland/fantasy-land#semigroup

?

README.md
joneshf: Ah, I hadn't even noticed that. Yes, `Card :: Type` is the correct thing here.

I guess that answers a question I had lingering in the back of my mind, "Are the signatures checked?"

README.md
boxofrox: I'm happy with that compromise.  At least now I know where to go for more information.

README.md
davidchambers: > `Card :: Type` is the correct thing here.

I'll update this.

> I guess that answers a question I had lingering in the back of my mind, "Are the signatures checked?"

I wish they were! I do appreciate the computer notifying me when I make a mistake.

README.md
davidchambers: According to [`S.type`][1],

> `'@@type'` properties should use the form `'<package-name>/<type-name>'`, where `<package-name>` is the name of the npm package in which the type is defined.

It's not clear which "package name" we should use. On one hand this type *is* defined in the sanctuary-def package (though only in the README). On the other hand, using `'sanctuary-def/Pair'` here may lead readers to assume that they should use the `sanctuary-def` namespace in their own projects.

Your `'my-app/Pair'` suggestion would avoid this confusion. Perhaps `'my-package/Pair'` is best of all.


[1]: https://github.com/plaid/sanctuary/blob/22828c682bda3847c58274da8719b59d5d258e47/index.js#L339-L357

README.md
davidchambers: Nice tweak, @CrossEye. Thanks!

:zap:

README.md

README.md
CrossEye: That sounds just right.  I hadn't seen the note on `S.type`.

README.md
```diff
@@ -0,0 +1,486 @@
...
+more general ones:
+
+```javascript
+//    inc :: Number -> Number
+const inc = add(1);
+
+inc(7);
+// => 8
+```
+
```
boxofrox: I'm not sure about this suggestion.  Is it worth mentioning that an error like `inc(7, 6)` will report an incorrect use of `add` instead of `inc`?

Also, would it be considered an acceptable practice to define `inc` as `const inc = def('inc', {}, [$.Number, $.Number], add(1));` if we really wanted specific function names in the error messages?  If not, it's certainly worth mentioning this as an anti-pattern.

This example is simple enough that the error should be obvious to figure out, but I'm imagining a situation where a newb dev, like myself, is reusing `inc` in a file far from its definition and getting an error unexpectedly about `add`.

boxofrox:
Overall I'm impressed with the API and utility of `def`.  I played with the `dc-everything-2` branch for awhile and the error messages are great, they tell me what the problem is and print the value so I know immediately what the code was looking at when the error resulted.  Bravo.

It's amazing how much functionality you've packed into this one function.  And besides the few comments I made against the README, it's perfect.  Well done, sir.

README.md
davidchambers: Excellent point, @boxofrox. Since none of us has much experience with the module at this point, it's too early to state a preference for one approach or the other. It's deserves to be mentioned in the README though. We could insert the following:

> It's worth noting that defining functions via partial application reduces the helpfulness of error messages:
>
> ```javascript
> inc('0');
> // ! TypeError: ‘add’ expected a value of type Number as its second argument; received "0"
> ```
>
> This can be avoided by using `def` to define the specialized function:
>
> ```javascript
> //    inc :: Number -> Number
> const inc = def('inc', {}, [$.Number, $.Number], add(1));
> ```

davidchambers:
Thank you for your feedback, @boxofrox. There's no substitute for requiring a module and taking it for a spin. Thanks for taking the time to do so. :)

README.md
boxofrox: :+1:

README.md
svozza: Maybe you could have the opening sentence look something like this:

```
sanctuary-def facilitates the definition of curried JavaScript functions which are explicit about the number of arguments to which they may be applied and the types of those arguments. sanctuary-def provides various default type definitions in what is termed an environment: this environment may also be augmented by user-defined types. For a more detailed description see <link to Environments section>.
```

davidchambers: Terrific! Let's do this. :)

index.js
```diff
@@ -0,0 +1,464 @@
...
+    return x != null && x['@@functional/placeholder'] === true;
+  };
+
+  //  quote :: String -> String
+  var quote = function(s) {
+    return '\u2018' + s + '\u2019';
+  };
+
+  //  forEach :: [a] -> (a -> Number -> Unit) -> Unit
+  var forEach = function(xs, f) {
```
svozza: Iteratee last? :open_mouth:

davidchambers: Yep! This is to enable some unattractive imperative code, and

```javascript
forEach(xs, function(x) {
  // [snip 50 lines]
});
```

is preferable to

```javascript
forEach(function(x) {
  // [snip 50 lines]
}, xs);
```

since one needn't scroll to the end of the block to see what it is that is being iterated.

svozza: Yep, makes sense.

package.json
```diff
@@ -0,0 +1,26 @@
...
+  "name": "sanctuary-def",
+  "version": "0.0.0",
+  "description": "Define strict, curried functions",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "git://github.com/plaid/sanctuary-def.git"
+  },
+  "dependencies": {
+    "ramda": "0.18.x"
```
svozza: I did a very rough calculation and sanctuary-def only uses about 20 per cent of the functions in Ramda, is the full dependency warranted? I know you don't like build processes but I'd like sanctuary-def to be as thin as possible (I have a project I contribute to in mind but I don't think the owner would consent to a large dependency like Ramda) so it would be nice to have an option of an ultraminimal version.

davidchambers: I agree! Ramda was helpful in getting this project up and running, but I'm now keen to remove the dependency. I plan to spend time this weekend sanding a few rough edges before merging this pull request and publishing the package. I'll remove the Ramda dependency while I'm at it. :)

svozza: Fantastic. This is really great work.

README.md
```diff
@@ -0,0 +1,486 @@
...
+  x => Object.prototype.toString.call(x) === '[object Number]' &&
+       Math.floor(x) === Number(x) &&
+       x >= Number.MIN_SAFE_INTEGER &&
+       x <= Number.MAX_SAFE_INTEGER
+);
+
+//    NonZeroInteger :: Type
+const NonZeroInteger = $.NullaryType(
+  'my-package/NonZeroInteger',
+  x => Integer.test(x) && Number(x) !== 0
```
svozza: This sort of comes out of nowhere. The only way you can figure out what exactly this `test` method does by inspecting the source code. I know it's fairly obvious but I still think the docs should be explicit.

boxofrox: Are you saying it's not clear that the predicate function--used to create the Integer type--becomes the `test` function?

svozza: Yes, just a comment in the example code to make it explicit where it's coming from would suffice.

boxofrox: This seems like a formal definition of functionality.  IMHO, it shouldn't be buried in an example, if for no other reason than I'm the kind of guy that overlooks those details in examples.  A better reason is this an API exposed to users, and likely (also imo) to be overlooked by distant-future changes.

I propose an alternative where the definition for nullary type is expanded to read...

> To define a nullary type `t` one must provide
> ...
> - a predicate which accepts any JavaScript value and returns `true` if
>   (and only if) the value is a member of `t`.  This predicate is publicly
>   exposed via a `test` method on the new nullary type (as demonstrated in the NonZeroInteger example).

That presents the `test` API up front for anyone supplementing nullary type with more functionality.

If acceptable, it's probably worth repeating this documentation for unary and binary types as well.  Thoughts?

svozza: Yes, very good point. I agree completely.

package.json
davidchambers: I've removed all uses of Ramda but one:

```javascript
var commonTypes = R.intersection(types, actualTypes);
```

This revealed something unfortunate: the result is not guaranteed to be a subset of *either* argument, due to the quirky equality semantics of Type values. Worse, the code *relies* on this behaviour. I'm not happy with this situation. Although I'm very much looking forward to using sanctuary-def in Sanctuary and in other projects, and I'm happy with its current API, I'm reluctant to merge an initial pull request containing code which is unclear to me even after spending dozens of hours with it. Thanksgiving will provide me an opportunity to rethink my approach. I'd like to provide the same API with dramatically simpler internals, to provide a solid foundation for future enhancements and to allow others to understand the code and contribute fixes.

boxofrox: No rush from this end.  Got plenty keeping me too busy to play more with JS + Ramda as it is.  Will have to wait to add Sanctuary-def to that mix anyway.  Enjoy the holiday :smile:

index.js
```diff
@@ -0,0 +1,693 @@
...
+          }
+        }
+        return typeVarMap;
+      } else {
+        return typeVarMap;
+      }
+    });
+  };
+
+  var doAllTheThings = function(o) {
```
svozza: Great function name. :laughing:

davidchambers:
I've pushed a big change I've been working on for the past week. The API has not changed, but I now understand the implementation! Ramda is no longer a dependency. :)

package.json
davidchambers: :ram: :zap:

index.js
davidchambers: It was a sign that I'd lost control of the code. The new implementation does not contain a function named `doAllTheThings`. :)

README.md
boxofrox: @davidchambers ping. Not sure if you disagreed or missed this conversation.

davidchambers: I agree. Making this change is on my mental to-do list. :)

benperez:
This looks really cool, nice job @davidchambers! I am looking forward to taking this for a test drive.

svozza:
I think you should just merge this and let people play around with it; the API can be ironed out on the back of some real world use then.

boxofrox:
You could adopt semantic versioning with major version 0 to indicate the API is subject to change until 1.0.0.

davidchambers:
I'm sorry that this pull request is taking longer to complete than I expected. I'm as eager as anybody to start using it in various projects. :)

I'm happy with the API, but there are still a few loose ends to tie up before I merge this and publish an initial version of the package. I've been working in parallel on a Sanctuary patch to add sanctuary-def as a dependency. This has been extremely useful as it has highlighted several issues that the test suite did not. I'll open a Sanctuary pull request *before* merging this one, since sanctuary-def must satisfy the current requirements of Sanctuary's type checking.

svozza:
Ah cool. I just thought you were being a perfectionist. :wink:

README.md
davidchambers: I made these changes:

```diff
--- a/README.md
+++ b/README.md
@@ -186,10 +186,10 @@ other exported types.
 
 To define a nullary type `t` one must provide:
 
-  - the name of `t`; and
+  - the name of `t` (exposed as `t.name`); and
 
   - a predicate which accepts any JavaScript value and returns `true` if
-    (and only if) the value is a member of `t`.
+    (and only if) the value is a member of `t` (exposed as `t.test`).
 
 ```haskell
 NullaryType :: String -> (* -> Boolean) -> Type
@@ -246,15 +246,16 @@ sum(['1', '2', '3', '4']);
 
 To define a unary type `t a` one must provide:
 
-  - the name of `t`;
+  - the name of `t` (exposed as `t.name`);
 
-  - a predicate which accepts any JavaScript value and returns `true` if
-    (and only if) the value is a member of `t x` for some type `x`;
+  - a predicate which accepts any JavaScript value and returns `true`
+    if (and only if) the value is a member of `t x` for some type `x`
+    (exposed as `t.test`);
 
-  - a function which takes any value of type `t a` and returns an array of
-    the values of type `a` contained in the `t`; and
+  - a function which takes any value of type `t a` and returns an array
+    of the values of type `a` contained in the `t` (exposed as `t._1`); and
 
-  - the type of `a`.
+  - the type of `a` (exposed as `t.$1`).
 
 ```haskell
 UnaryType :: String -> (* -> Boolean) -> (t a -> [a]) -> Type -> Type
@@ -307,20 +308,21 @@ fromMaybe(0, Just('XXX'));
 
 To define a binary type `t a b` one must provide:
 
-  - the name of `t`;
+  - the name of `t` (exposed as `t.name`);
 
-  - a predicate which accepts any JavaScript value and returns `true` if
-    (and only if) the value is a member of `t x y` for some types `x` and `y`;
+  - a predicate which accepts any JavaScript value and returns `true`
+    if (and only if) the value is a member of `t x y` for some types
+    `x` and `y` (exposed as `t.test`);
 
-  - a function which takes any value of type `t a b` and returns an array of
-    the values of type `a` contained in the `t`;
+  - a function which takes any value of type `t a b` and returns an array
+    of the values of type `a` contained in the `t` (exposed as `t._1`);
 
-  - a function which takes any value of type `t a b` and returns an array of
-    the values of type `b` contained in the `t`;
+  - a function which takes any value of type `t a b` and returns an array
+    of the values of type `b` contained in the `t` (exposed as `t._2`);
 
-  - the type of `a`; and
+  - the type of `a` (exposed as `t.$1`); and
 
-  - the type of `b`.
+  - the type of `b` (exposed as `t.$2`).
 
 ```haskell
 BinaryType :: String -> (* -> Boolean) -> (t a b -> [a]) -> (t a b -> [b]) -> Type -> Type -> Type
```

Thanks for catching this omission. It's good to be documenting `_1`, `_2`, `$1`, and `$2` as well!

boxofrox: Very nice.

index.js
```diff
@@ -0,0 +1,693 @@
...
+        }
+        var values = _values.slice();
+        var indexes = [];
+        for (var i = 0; i < _indexes.length; i += 1) {
+          var index = _indexes[i];
+
+          if (i < arguments.length &&
+              !(typeof arguments[i] === 'object' &&
+                arguments[i] != null &&
+                arguments[i]['@@functional/placeholder'] === true)) {
```
albrow: I wonder if the placeholder could itself be defined as a `NullaryType`? Something like:

``` javascript
const Placeholder = $.NullaryType(
  'sanctuary-def/Placeholder',
  x => typeof x === 'object' &&
    x != null &&
    x['@@functional/placeholder'] === true
);
```

Do you think that would lend any clarity or would you prefer to not define any types within sanctuary-def itself?

davidchambers: Interesting idea! I'm not sure it's a good idea, though. A placeholder represents the absence of a value, so treating a placeholder as a value could be a source of confusion. Also, we'd still need to check for a placeholder here explicitly, so we wouldn't actually leverage the existing type-checking machinery.

README.md
```diff
@@ -0,0 +1,489 @@
...
+const $ = require('sanctuary-def');
+```
+
+The next step is to apply `$.create` to a list of types (an environment):
+
+```javascript
+const def = $.create($.env);
+```
+
+`$.env` is a list of types. One could augment it with custom types
```
albrow: After reading this the first time, it was not clear to me why one would want to augment `$.env` with custom types. In the following examples, it appeared that you were declaring custom types, such as `Maybe`, `$Pair`, `Rank`, and `Suit` and then just passing them directly as arguments to `def`.

After reading again carefully and taking a look at the source code, I think the way it actually works is that you have to define all your types beforehand and pass them to `$.create` as part of the environment. Any type that you pass to `def` which is not part of the environment will not work. Is that correct? If so, I think you should clarify this in the section about the `$.create` function and consider adding an explicit call to `$.create` in the later example code.

README.md
```diff
@@ -0,0 +1,489 @@
...
+```javascript
+const $ = require('sanctuary-def');
+```
+
+The next step is to apply `$.create` to a list of types (an environment):
+
+```javascript
+const def = $.create($.env);
+```
+
```
albrow: You have included a list of all the types that are part of `$.env` further down the page (at which point there is actually no mention of `$.env`). I would prefer to see the types enumerated here, since one of my first questions after reading this sentence was "What types are included in `$.env`?".

If the list of types defined in `$.env` becomes to large, it may be appropriate to link to a full list of types further down the page or in a separate page, rather than trying to list them all right here.

index.js
```diff
@@ -0,0 +1,745 @@
...
+      var expArgTypes = expTypes.slice(0, -1);
+      var arity = expArgTypes.length;
+      if (arity > 9) {
+        throw new RangeError(
+          LEFT_SINGLE_QUOTATION_MARK + 'def' + RIGHT_SINGLE_QUOTATION_MARK +
+          ' cannot define a function with arity greater than nine'
+        );
+      }
+
+      assertExpectedTypesInEnvironment(name)(expTypes);
```
davidchambers: On Friday, @albrow raised an excellent point while reviewing the README. He was surprised to discover that it's possible to use a type in a function definition *without* first including that type in the environment. This is because the environment is only important when type variables are involved. When one applies `a -> a -> a` to `42`, sanctuary-def must determine the types of which `42` is a member. This is done by filtering the list of types provided to `$.create`.

Alex and I agreed that it's confusing to require types to be included in the environment in just *some* cases. Rather than document the actual behaviour, we decided it's preferable to state that `def` only accepts types in its environment. We then decided to go a step further and make it a run-time error to reference a "foreign" type in a function definition.

README.md
davidchambers: I've reworked the introduction. I believe the following paragraph captures the meaning conveyed by your suggested addition, Stefano:

> The next step is to define an environment. An environment is a list of [types](https://github.com/plaid/sanctuary-def/tree/dc-everything-2#types) such as `$.Number` and `$.String`. `$.env` is a default environment, which may be augmented with custom types:

```javascript
//    Integer :: Type
const Integer = ...;

//    NonZeroInteger :: Type
const NonZeroInteger = ...;

//    env :: [Type]
const env = $.env.concat([Integer, NonZeroInteger]);
```

davidchambers:
:guardsman:

svozza:
Fantastic work, David! Looking forward to taking this for a spin myself.

PR #9, opened 2016-01-03
support "nullable" types

### LINKED ISSUE #8: support "nullable" types
`null` is one of the unfortunate realities of JavaScript. In order to be useful in a broad range of settings, sanctuary-def should acknowledge this and enable the definition of types which include `null` as a member. `displayName :: Nullable String`, for example, states that `displayName` is either `null` or a string value.

It's currently possible to define `NullableString` via `$.NullaryType`:

```javascript
//    NullableString :: Type
const NullableString = $.NullaryType(
  'my-package/NullableString',
  x => x === null || $.String.test(x)
);

const def = $.create($.env.concat([NullableString]));

//    f :: NullableString -> NullableString
const f = def('f', {}, [NullableString, NullableString], x => x);
```

What about polymorphism? In Sanctuary, for example, we'd like [`S.toMaybe`][1] to have type `Nullable a -> Maybe a`. At first glance we already have a workable solution:

```javascript
//    Nullable :: Type -> Type
const Nullable = $.UnaryType(
  'my-package/Nullable',
  x => true,
  nullable => nullable === null ? [] : [nullable]
);
```

If `Nullable` exists in the environment, what are the types of which `'foo'` is a member? `String`, of course, but also `Nullable _` for some `_`. Resolving `_` is infinitely recursive because the "inner" value of a nullable is the nullable itself. We determine that `'foo' :: String` and `'foo' :: Nullable String`, but we also determine that `'foo' :: Nullable (Nullable String)`, `'foo' :: Nullable (Nullable (Nullable String))`, and so on.

sanctuary-def should provide a special `$.Nullable` type constructor.

[1]: https://github.com/plaid/sanctuary#toMaybe

davidchambers:
Closes #8

### DIFF
--- a/README.md
+++ b/README.md
@@ -433,6 +433,38 @@ dist(0);
 // ! TypeError: ‘dist’ expected a value of type { x :: FiniteNumber, y :: FiniteNumber } as its first argument; received 0
 ```
 
+#### `Nullable`
+
+`Nullable` is used to construct types which include `null` as a member.
+
+To define `t :: Nullable a` one must provide:
+
+  - the type of `a` (exposed as `t.$1`).
+
+```haskell
+Nullable :: Type -> Type
+```
+
+For example:
+
+```javascript
+//    toUpper :: Nullable String -> Nullable String
+const toUpper =
+def('toUpper',
+    {},
+    [$.Nullable($.String), $.Nullable($.String)],
+    nullable => nullable === null ? null : nullable.toUpperCase());
+
+toUpper(null);
+// => null
+
+toUpper('abc');
+// => 'ABC'
+
+toUpper(['abc']);
+// ! TypeError: ‘toUpper’ expected a value of type (Nullable String) as its first argument; received ["abc"]
+```
+
 ### Type classes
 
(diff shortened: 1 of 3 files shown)


svozza:
LGTM.

davidchambers:
I must document this function before merging this pull request.

davidchambers:
@svozza, does the README addition read well to you?

README.md
```diff
@@ -433,6 +433,38 @@ dist(0);
...
+
+For example:
+
+```javascript
+//    toUpper :: Nullable String -> Nullable String
+const toUpper =
+def('toUpper',
+    {},
+    [$.Nullable($.String), $.Nullable($.String)],
+    R.unless(R.equals(null), R.toUpper));
```
svozza: Do we want the docs for `sanctuary-def` to assume a deep familiarity with Ramda? I use it every day and I still had to look up what `R.unless` does just to be sure I was reading the example properly.

davidchambers: Great point! Let's use a lambda instead.

svozza: :+1:

Issue #11, opened 2016-01-03
Should $.Date Check If Date Is Valid?


svozza:
I've been playing around with `sanctuary-def` this morning and encountered a strange issue when working with `Date` types that was quite hard to debug. I was getting these strange `RangeError`s being thrown and turns out it's because the `_show` function uses `Date#toIsoString()`, which for invalid Date objects throws this exception. 

I found this when creating my own `ValidDate` type:

```js
$.NullaryType(
    'sanctuary-date-poc/ValidDate',
    date => Object.prototype.toString.call(date) === "[object Date]" &&
        date.getTime() === date.getTime()
);
```

But it occurs to me that an invalid date is really a type error so maybe `sanctuary-def` could enforce this constraint out of the box.

davidchambers:
`InvalidDate` *is* a member of the Date type, though I agree it's not a particularly useful one. I'd like `$.Date` to remain as is, but I'm totally on board with defining `$.ValidDate`.

I can think of several other types we could provide:

  - `$.ValidNumber` (`$.Number` sans `NaN`);
  - `$.FiniteNumber` (`$.Number` sans `Infinity`, `-Infinity`, and `NaN`);
  - `$.NonZeroValidNumber` (`$.ValidNumber` sans `0` and `-0`);
  - `$.NonZeroFiniteNumber` (`$.FiniteNumber` sans `0` and `-0`);
  - `$.Integer`;
  - `$.NonZeroInteger`; and
  - `$.RegexFlags` (`''`, `'g'`, `'i'`, `'m'`, `'gi'`, `'gm'`, `'im'`, and `'gim'`).

I'm nervous about types such as `$.PositiveNumber` and `$.NegativeInteger` given the existence of both `0` and `-0` in JavaScript. It's unclear whether `0` should be considered a positive number in this context.

> [T]he `_show` function uses `Date#toIsoString()`, which for invalid Date objects throws this exception.

This is certainly a bug, which should be fixed here and in Ramda.

svozza:
I'm happy to put together a PR for ValidDate as it's very useful for me
right now. Do you want me to fix the toString issue while I'm at it or do
you want to do it in a separate PR?

davidchambers:
> I'm happy to put together a PR for ValidDate as it's very useful for me right now.

Terrific! Please do. :)

> Do you want me to fix the toString issue while I'm at it or do you want to do it in a separate PR?

I'm about to open a pull request for that fix. :)

[...]

PR #12, opened 2016-01-03
handle "Invalid Date" in show

davidchambers:
See #11, ramda/ramda#1576

We should wait until the Ramda pull request is accepted before merging this pull request.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "jscs": "2.7.x",
     "jshint": "2.8.x",
     "mocha": "2.x.x",
-    "ramda": "0.18.x",
+    "ramda": "~0.19.1",
     "xyz": "0.5.x"
   },
   "files": [
--- a/index.js
+++ b/index.js
@@ -157,7 +157,9 @@
           'new Boolean(' + recur(x.valueOf()) + ')' :
           x.toString();
       case 'Date':
-        return 'new Date(' + quote(x.toISOString()) + ')';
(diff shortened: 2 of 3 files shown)


svozza:
I'm guessing it's safe to use the global `isNaN` function here because the type coercion it does isn't a problem here.

davidchambers:
> I'm guessing it's safe to use the global `isNaN` function because the type coercion it does isn't a problem here, i.e, `valueOf` always returns a number.

Yes. Initially I used `x.valueOf() === x.valueOf()` as the check, but I then remembered that `isNaN` is well behaved when the domain is restricted to values of type `Number`. `isNaN` is a bit clearer in this case.

I just updated this pull request to depend on `ramda@~0.19.1`, so the test for `R.toString`-like string representations is now a one-liner. :)

package.json
```diff
@@ -13,7 +13,7 @@
     "jscs": "2.7.x",
     "jshint": "2.8.x",
     "mocha": "2.x.x",
-    "ramda": "0.18.x",
+    "ramda": "~0.19.1",
```
albrow: Is there a reason we used the `~` syntax instead of `0.19.x`?

davidchambers: Yes. In this case we're depending on behaviour which changed between v0.19.0 and v0.19.1, via ramda/ramda#1576. We need to use at least v0.19.1 to have the "uses `R.toString`-like string representations" tests pass.

PR #13, opened 2016-01-04
Add ValidDate type

svozza:
The tests will fail until #12 is merged but I figured I may as well raise the PR because I'm sure there are other changes I'll need to make too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,6 +232,14 @@ $.Undefined :: Type
 
 Type whose sole member is `undefined`.
 
+#### `ValidDate`
+
+```haskell
+$.ValidDate :: Type
+```
+
+Type comprising every [`Date`](#date) value except `new Date(NaN)`.
+
 ### `env`
 
 `$.env` is a list of [types](#types):
--- a/index.js
+++ b/index.js
@@ -356,6 +356,12 @@
     ($.Undefined  = type0('Undefined'))
   ];
 
+  //  ValidDate :: Type
+  $.ValidDate = $.NullaryType(
+    'sanctuary-def/ValidDate',
+    function(x) { return $.Date.test(x) && !isNaN(x.valueOf()); }
+  );
+
   //  arity :: (Number, Function) -> Function
   var arity = function(n, f) {
     return (
--- a/test/index.js
+++ b/test/index.js
@@ -545,6 +545,23 @@ describe('def', function() {
                    'as its first argument; received ["abc"]'));
   });
 
+  it('supports the "ValidDate" type', function() {
+    var def = $.create($.env.concat([$.ValidDate]));
+
+    //  sinceEpoch :: ValidDate -> Number
+    var sinceEpoch = def('sinceEpoch',
+                         {},
+                         [$.ValidDate, $.Number],
+                         function(date) { return date.getTime() / 1000; });
+
+    throws(function() { sinceEpoch(new Date('foo')); },
+           errorEq(TypeError,
+                   '‘sinceEpoch’ expected a value of type ValidDate ' +
+                   'as its first argument; received new Date(NaN)'));
+
+    eq(sinceEpoch(new Date(123456)), 123.456);
+  });
+
   it('uses R.toString-like string representations', function() {
     //  f :: Null -> Null
     var f = def('f', {}, [$.Null, $.Null], function(x) { return x; });


index.js
```diff
@@ -306,6 +306,11 @@
     };
   };
 
+  var ValidDate = $.NullaryType(
+    'ValidDate',
```
davidchambers: This should be `'sanctuary-def/ValidDate'`. :)

index.js
```diff
@@ -306,6 +306,11 @@
     };
   };
 
+  var ValidDate = $.NullaryType(
```
davidchambers: Let's include `ValidDate :: Type`.

index.js
```diff
@@ -306,6 +306,11 @@
...
+    'ValidDate',
+    function (x) { return $$typeEq('Date')(x) && !isNaN(x.getTime()) }
```
davidchambers: For some reason I prefer to use `.valueOf()` rather than `.getTime()`. I can't justify this preference. ;)

test/index.js
```diff
@@ -528,6 +529,19 @@ describe('def', function() {
     eq(id([{x: 0, y: 0}, {x: 1, y: 1}]), [{x: 0, y: 0}, {x: 1, y: 1}]);
   });
 
+  it('supports "ValidDate" types', function () {
```
davidchambers: Since there's just one ValidDate type, let's `s/types/type/`.

test/index.js
```diff
@@ -528,6 +529,19 @@ describe('def', function() {
...
+    var sinceEpoch = def('sinceEpoch', {}, [$.ValidDate, $.Number], function(date) {
+      return date.getTime() / 1000 });
```
davidchambers: I'd prefer this formatting:

```javascript
//  sinceEpoch :: ValidDate -> Number
var sinceEpoch =
def('sinceEpoch',
    {},
    [$.ValidDate, $.Number],
    function(date) { return date.getTime() / 1000; });
```

test/index.js
```diff
@@ -528,6 +529,19 @@ describe('def', function() {
...
+
+    throws(function() { sinceEpoch(new Date('foo')); },
+      errorEq(TypeError,
+        '‘sinceEpoch’ expected a value of type ValidDate ' +
+        'as its first argument; received new Date(NaN)'));
```
davidchambers: :triangular_ruler:

test/index.js
```diff
@@ -528,6 +529,19 @@ describe('def', function() {
...
+
+    eq(sinceEpoch(new Date(Date.UTC(1973, 11, 28))), 125884800);
+
```
davidchambers: :space_invader:

index.js
svozza: Yeah, the reason I used `getTime` was because I read in the Mozilla [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/valueOf):

>This method is usually called internally by JavaScript and not explicitly in code.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/valueOf

But I'm happy to change it.

index.js

index.js

test/index.js

test/index.js
svozza: :zap:  What editor are you using? Do you have a way of exporting your formatting rules? I'm always messing up when it comes to this sort of thing.

test/index.js

svozza:
This is really weird, the tests are failing for me locally on `uses R.toString-like string representations`.

svozza:
Oops, I hadn't upgraded to the new version of Ramda.

index.js
davidchambers: That's a solid argument in favour of `.getTime()`. Since we used `.valueOf()` in #12, though, I'd prefer to use it here as well.

davidchambers:
I don't think we should include `$.ValidDate` in `$.env`. When one applies a function of type `a -> a -> a`, we determine all the types of which the first argument is a member. We then do the same for the second argument. If the intersection of these two sets is empty, we throw a type error _which lists all the first argument's types_. `1` could soon belong to half a dozen types. To avoid long error messages I suggest we leave `$.env` as is and make future types available à la carte.

We need to work out how best to document additional types such as `$.ValidDate`. Mind having a go at updating the README?

index.js
svozza: I had a feeling you'd say that. :)

:zap:

svozza:
Yeah, I wasn't sure where to put it so I just stuck it in `env`. I'll move it out now and update the README.

index.js
```diff
@@ -306,6 +306,12 @@
     };
   };
 
+  //  ValidDate :: Type
+  var ValidDate = $.NullaryType(
+    'sanctuary-def/ValidDate',
+    function(x) { return $$typeEq('Date')(x) && !isNaN(x.valueOf()); }
+  );
+
```
davidchambers: Were we to define this after `$.Date` we could leverage `$.Date.test`:

```javascript
//  ValidDate :: Type
$.ValidDate = $.NullaryType(
  'sanctuary-def/ValidDate',
  function(x) { return $.Date.test(x) && !isNaN(x.valueOf()); }
);
```

This would make it very clear that `$.ValidDate` is `$.Date` sans `new Date(NaN)`.

test/index.js
```diff
@@ -528,6 +529,20 @@ describe('def', function() {
     eq(id([{x: 0, y: 0}, {x: 1, y: 1}]), [{x: 0, y: 0}, {x: 1, y: 1}]);
   });
 
+  it('supports the "ValidDate" type', function() {
+    var sinceEpoch = def('sinceEpoch',
```
davidchambers: Let's add `sinceEpoch :: ValidDate -> Number`.

svozza:
So I've added documentation about `ValidDate`. When we add more types, this should become a section that lists the various additional types available and gives examples of their use but for the moment I though it would look odd for just one type.

svozza:
I envisage the final documentation looking like this:

### Types

sanctuary-def defines `Type` values for JavaScript's built-in types:

  - `$.Array :: Type -> Type`
  - `$.Boolean :: Type`
  - `$.Date :: Type`
  - `$.Error :: Type`
  - `$.Function :: Type`
  - `$.Null :: Type`
  - `$.Number :: Type`
  - `$.Object :: Type`
  - `$.RegExp :: Type`
  - `$.String :: Type`
  - `$.Undefined :: Type`

sanctuary-def also defines `$.Any`, the type of which every JavaScript value
is a member.

In addition, many more restrictive versions of the above types are available but not
loaded into the environment by default. These include:

- `$.ValidDate`  
- `$.ValidNumber`
- `$.FiniteNumber`
- `$.NonZeroValidNumber`
- `$.NonZeroFiniteNumber`
- `$.Integer`;
- `$.NonZeroInteger`
- `$.RegexFlags`

#### `ValidDate`
Defines a date type that does not include invalid JS date objects.

```js
const def = $.create($.env.concat([$.ValidDate]));

const sinceEpoch = def('sinceEpoch',
                       {},
                       $.ValidDate, $.Number],
                       date => date.getTime() / 1000);

sinceEpoch(new Date('foo'))
// ! TypeError: ‘sinceEpoch’ expected a value of type ValidDate as its first argument; 
// received new Date(NaN)
```

And so on...

davidchambers:
This looks good, though there's another structure which would involve less duplication:

```markdown
### Types
#### `Any`
#### `Array`
#### `Boolean`
#### `Date`
#### `Error`
#### `FiniteNumber`
#### `Function`
#### `Integer`
#### `NonZeroFiniteNumber`
#### `NonZeroInteger`
#### `NonZeroValidNumber`
#### `Null`
#### `Number`
#### `Object`
#### `RegExp`
#### `RegexFlags`
#### `String`
#### `Undefined`
#### `ValidDate`
#### `ValidNumber`
### `env`
```

The __`env`__ section would simply list the types included in `$.env`, each linking to its documentation.

I'll put together a pull request to restructure the documentation, so this pull request can remain focused on defining `$.ValidDate`. :)

svozza:
Will I remove the interim documentation I've added in this PR?

davidchambers:
You can update this pull request now, Stefano. :)

svozza:
Great stuff! :zap:

README.md
```diff
@@ -232,6 +232,14 @@ $.Undefined :: Type
 
 Type whose sole member is `undefined`.
 
+#### `ValidDate`
+
+```haskel
```
davidchambers: `s/haskel/haskell/`

README.md
```diff
@@ -232,6 +232,14 @@ $.Undefined :: Type
...
+$.ValidDate :: Type
+```
+
+Type comprising all valid JavaScript date objects.
```
davidchambers: For consistency with #18, I suggest:

```markdown
Type comprising every [`Date`](#date) value except `new Date(NaN)`.
```

test/index.js
```diff
@@ -545,6 +545,23 @@ describe('def', function() {
...
+                         {},
+                         [$.ValidDate, $.Number],
+                         function(date) { return date.getTime() / 1000; });
+
+    throws(function() { sinceEpoch(new Date('foo')); },
+           errorEq(TypeError,
+                   '‘sinceEpoch’ expected a value of type ValidDate ' +
+                   'as its first argument; received new Date(NaN)'));
+
+    eq(sinceEpoch(new Date(Date.UTC(1973, 11, 28))), 125884800);
```
davidchambers: Perhaps a simpler assertion would be clearer?

```javascript
eq(sinceEpoch(new Date(123456)), 123.456);
```

README.md
svozza: :zap

README.md

test/index.js

test/index.js
```diff
@@ -545,6 +545,23 @@ describe('def', function() {
                    'as its first argument; received ["abc"]'));
   });
 
+  it('supports the "ValidDate" type', function() {
+    var def = $.create($.env.concat([$.ValidDate]));
+
+    // sinceEpoch :: ValidDate -> Number
```
davidchambers: Nit: Let's align `sinceEpoch` with the line below.

davidchambers:
:deciduous_tree:

PR #14, opened 2016-01-05
readme: restructure "types" section

davidchambers:
This structure will accommodate new types such as `$.ValidDate` to be added in #13.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,8 +11,10 @@ const $ = require('sanctuary-def');
 ```
 
 The next step is to define an environment. An environment is a list of
-[types](#types) such as `$.Number` and `$.String`. `$.env` is a default
-environment, which may be augmented with custom types:
+[types](#types) such as `$.Number` and `$.String`. [`$.env`](#env) is an
+environment containing all the built-in JavaScript types. It may be used
+as the basis for environments which include custom types in addition to
+the built-in types:
 
 ```javascript
 //    Integer :: Type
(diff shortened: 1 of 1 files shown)


svozza:
Excellent. I think the way I put an example in my docs was probably overkill and a short sentence - like you've done here - describing the type will suffice.

README.md
```diff
@@ -124,22 +124,123 @@ add('X');
...
+
+Type whose sole member is `null`.
+
+#### `Number`
+
+```haskell
+$.Number :: Type
+```
+
+Type comprising every Number value (including Number objects).
```
svozza: Maybe note that `NaN` is a number too.

README.md
```diff
@@ -124,22 +124,123 @@ add('X');
...
+
+#### `Object`
+
+```haskell
+$.Object :: Type
+```
+
+Type comprising every Object value, including values of more specific
```
svozza: Maybe mention that `null` is an object too.

README.md

README.md
davidchambers: I realized that this description is incorrect. `$.Object` is a type comprising every value satisfying `Object.prototype.toString.call(x) === '[object Object]'`. I'll update the description.

svozza: Yes, makes sense.

svozza:
LGTM.

README.md
```diff
@@ -11,8 +11,8 @@ const $ = require('sanctuary-def');
 ```
 
 The next step is to define an environment. An environment is a list of
-[types](#types) such as `$.Number` and `$.String`. `$.env` is a default
-environment, which may be augmented with custom types:
+[types](#types) such as `$.Number` and `$.String`. [`$.env`](#env) is a
+default environment, which may be augmented with custom types:
```
wmrowan: A default environment or the default environment? Are there other possible default environments?

davidchambers: Good question. When I see _the default value is `42`_ I assume that if I don't provide a value, `42` will be used. In this case, omitting the environment argument is not an option. `$.env` is simply a value which is near at hand which can be provided as the environment argument. It's not really a default at all, I suppose. Perhaps we should simply write _`$.env` is one possible environment_. What do you think?

wmrowan: I like that but would add further clarity. Most users of sanctuary-def will want to use the technique demonstrated in the example to define environments that extend `$.env`. How's this:

_The next step is to define an environment. An environment is a list of types such as `$.Number` and `$.String`. An environment containing all the builtin JavaScript types can be found at `$.env`. It can be used as the basis for environments that include both the builtin types and custom types:_

README.md
```diff
@@ -124,22 +124,124 @@ add('X');
 
 ### Types
 
-sanctuary-def defines `Type` values for JavaScript's built-in types:
+Conceptually, a type is a set of values. One can think of a value of
+type `Type` as a function of type `Any -> Boolean` (though this is an
```
wmrowan: I might clarify this by adding something like `One can think of a value of type `Type` as a function of type `Any -> Boolean` that tests values for membership in this set.`

README.md
```diff
@@ -124,22 +124,124 @@ add('X');
...
+
+Type whose sole member is `null`.
+
+#### `Number`
+
+```haskell
+$.Number :: Type
+```
+
+Type comprising every Number value (including `NaN` and Number objects).
```
wmrowan: This statement is oxymoronic.

README.md
```diff
@@ -124,22 +124,124 @@ add('X');
...
+
+#### `Object`
+
+```haskell
+$.Object :: Type
+```
+
+Type comprising every "plain" Object value; specifically, every `x` satisfying
+`Object.prototype.toString.call(x) === '[object Object]'`.
```
wmrowan: Could you clarify this further? I'm not familiar with the term "plain object". What is this test meant to distinguish?

davidchambers: We want to say that `$.Object` is more restrictive than the type comprising every `x` satisfying `x instanceof Object`. `[]`, for example, is a member of this broader type but is not a member of `$.Object`. The term "plain object" is used in lodash: [`_.isPlainObject`][1].

There may well be a better way to word this paragraph.


[1]: https://lodash.com/docs#isPlainObject

README.md
davidchambers: I like it! I'll make this change.

README.md
davidchambers: Very nice!

README.md
davidchambers: Are you trolling? :stuck_out_tongue_winking_eye:

README.md

README.md

README.md
svozza: :laughing:

README.md
svozza: Would it be correct to write something like:

> Type comprising every "plain" Object value, specifically any object created with `new Object()`, `Object.create` or the object literal syntax.

wmrowan: I like @svozza's wording.

davidchambers: :zap: I added a slightly modified version of Stefano's proposal.

wmrowan:
:earth_americas:

README.md
```diff
@@ -124,22 +126,128 @@ add('X');
...
+#### `Object`
+
+```haskell
+$.Object :: Type
+```
+
+Type comprising every "plain" Object value. Specifically, values created via:
+
+  - object literal syntax;
+  - `Object.create(null)`; or
```
svozza: What about `Object.create(someObject)`? E.g.,

```js
Object.prototype.toString.call(Object.create(new Array()));
// => [object Object]
```

davidchambers: Good point. This shows how little I've used `Object.create`. Shall we simply replace "`Object.create(null)`" with "[`Object.create`][1]"?


[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create

svozza: Yep, sounds good. Think this PR is good to go now.

davidchambers:
Thank you both for your excellent feedback! :icecream:

PR #17, opened 2016-01-09
support polymorphic Nullable types

davidchambers:
#9 did not support types such as `Nullable a -> Nullable a` and `a -> Nullable a -> a`. This change addresses the shortcoming.

The new implementation defines `$.Nullable` via `$.UnaryType`, as this turned out to be simpler. This necessitated the addition of a `t.name === 'sanctuary-def/Nullable'` check, which is unfortunate. I haven't thought of a simpler way to avoid `(Nullable (Nullable (Nullable ...)))` infinite recursion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,14 @@ $.Null :: Type
 
 Type whose sole member is `null`.
 
+#### `Nullable`
+
+```haskell
+$.Nullable :: Type -> Type
+```
+
+Constructor for types which include `null` as a member.
+
 #### `Number`
 
 ```haskell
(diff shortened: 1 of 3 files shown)


davidchambers:
I'd love a review from @albrow, @svozza, @wmrowan, or anyone else who may be interested.

svozza:
Do we want to test what happens with nested `Nullable` types anyway? Just in case the implementation changes.

davidchambers:
> Do we want to test what happens with nested `Nullable` types anyway?

Are you suggesting a user might write `$.Nullable($.Nullable($.String))`? :trollface:

svozza:
Hehe. Never underestimate the insanity of your users! :stuck_out_tongue_winking_eye: What I meant was that if there is code there to guard against this then we should probably have a test for it.

davidchambers:
We don't have logic to prevent someone from defining `Nullable (Nullable String)`, though it's unclear what this even means. We do have logic for preventing `determineActualTypes` from recursing infinitely, and this logic is tested.

svozza:
Ah, fair enough. My mistake.

PR #19, opened 2016-01-11
Add linting to Makefile test target

svozza:
A few times I've forgotten to run the `lint` target before committing resulting in failed build so I think that if something can fail the build then it's natural home is the `test` target. 

I've also added an `npm test` script to `package.json`. I know you dislike aliases David so I'm happy to remove it, it's just that every other project I work on uses `npm test` to kick off tests and I keep accidentally typing it into my terminal only to get an error.


davidchambers:
> I know you dislike aliases David so I'm happy to remove it, it's just that every other project I work on uses `npm test` to kick off tests and I keep accidentally typing it into my terminal only to get an error.

I'm sympathetic. I'm so used to `make setup`, `make test`, and `make lint` that I get tripped when working on projects which don't support these. I'm happy to support `npm test`.

I'd like to keep `make test` and `make lint` separate. When working on a patch I like to run `make test`; only once I'm ready to commit will I run `make lint` to check my style. As a compromise, we could make `npm test` an alias for `make test lint`. :)

svozza:
Sounds good!

PR #20, opened 2016-01-12
Add Enum Type

### LINKED ISSUE #15: Enum Type
I have a function that takes an argument that is a string but I want to restrict the values that string can be to 5 distinct strings. I created this type:

```js
const $TimeUnit = $.NullaryType(
    'sanctuary-date-poc/TimeUnit',
    x => typeof x === 'string' && ['milliseconds',
    'seconds',
    'minutes',
    'hours',
    'days'].indexOf(x) !== -1
);
```

The problem here is that it would be nice to have the error message print out the allowed values when it's passed one that's incorrect. Currently all that comes back is the name of the type. Is there a better way to do this? Or is there a case to be made for a `$.Enum` type?

svozza:
Closes #15. Wasn't too sure about the String representation so I followed the example of `Nullable`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -542,6 +542,47 @@ showCard(Pair('X', '♠'));
 // ! TypeError: ‘showCard’ expected a value of type (Pair Rank Suit) as its first argument; received Pair("X", "♠")
 ```
 
+#### `EnumType`
+
+`EnumType` is used to construct [enumerated types][5].
+
+To define an enumerated type one must provide:
+
+  - an array of values with distinct [`R.toString`][6] representations.
+
+```haskell
+EnumType :: [Any] -> Type
+```
+
+For example:
+
+```javascript
+//    TimeUnit :: Type
+const TimeUnit = $.EnumType(['milliseconds', 'seconds', 'minutes', 'hours']);
+
+const def = $.create($.env.concat([TimeUnit, $.ValidDate, $.ValidNumber]));
+
+//    convertTo :: TimeUnit -> ValidDate -> ValidNumber
+const convertTo =
+def('convertTo',
+    {},
+    [TimeUnit, $.ValidDate, $.ValidNumber],
+    function recur(unit, date) {
+      switch (unit) {
+        case 'milliseconds': return date.valueOf();
+        case 'seconds':      return recur('milliseconds', date) / 1000;
+        case 'minutes':      return recur('seconds', date) / 60;
+        case 'hours':        return recur('minutes', date) / 60;
+      }
+    });
+    
+convertTo('seconds', new Date(1000));
+// => 1
+
+convertTo('days', new Date(1000));
+// ! TypeError: ‘convertTo’ expected a value of type ("milliseconds" | "seconds" | "minutes" | "hours") as its first argument; received "days"
+```
+
 #### `RecordType`
 
 `RecordType` is used to construct record types. The type definition specifies
@@ -642,8 +683,8 @@ _concat([1, 2], 'buzz');
 
 The type of `_concat` is misleading: it suggests that it can operate on any
 two values of *any* one type. In fact there's an implicit constraint, since
-the type must support concatenation (in [mathematical][5] terms, the type
-must have a [semigroup][6]). The run-time type errors that result when this
+the type must support concatenation (in [mathematical][7] terms, the type
+must have a [semigroup][8]). The run-time type errors that result when this
 constraint is violated are not particularly descriptive:
 
 ```javascript
@@ -686,5 +727,7 @@ Multiple constraints may be placed on a type variable by including multiple
 [2]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SyntaxError
 [3]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError
 [4]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create
-[5]: https://en.wikipedia.org/wiki/Semigroup
-[6]: https://github.com/fantasyland/fantasy-land#semigroup
+[5]: https://en.wikipedia.org/wiki/Enumerated_type
+[6]: http://ramdajs.com/docs/#toString
+[7]: https://en.wikipedia.org/wiki/Semigroup
+[8]: https://github.com/fantasyland/fantasy-land#semigroup
--- a/index.js
+++ b/index.js
@@ -280,6 +280,22 @@
     });
   };
 
+  //  EnumType :: [Any] -> Type
+  $.EnumType = function(members) {
+    var reprs = map(members, show);
+    return {
+      type: 'ENUM',
+      test: function(x) {
+        var repr = show(x);
+        for (var idx = 0; idx < reprs.length; idx += 1) {
+          if (reprs[idx] === repr) return true;
+        }
+        return false;
+      },
+      toString: always('(' + reprs.join(' | ') + ')')
+    };
+  };
+
   //  RecordType :: {Type} -> Type
   var RecordType = $.RecordType = function(fields) {
     return {
@@ -494,6 +510,8 @@
         return t1.type === t2.type && t1.name === t2.name &&
                equalTypes(t1.$1, t2.$1) &&
                equalTypes(t1.$2, t2.$2);
+      case 'ENUM':
+        return t1.type === t2.type && show(t1) === show(t2);
       case 'RECORD':
         return t1.type === t2.type && show(t1) === show(t2);
       /* istanbul ignore next */
--- a/test/index.js
+++ b/test/index.js
@@ -464,6 +464,64 @@ describe('def', function() {
                    'as its first argument; received 0.5'));
   });
 
+  it('supports enumerated types', function() {
+    //  TimeUnit :: Type
+    var TimeUnit = $.EnumType(['milliseconds', 'seconds', 'minutes', 'hours']);
+
+    var def = $.create($.env.concat([TimeUnit, $.ValidDate, $.ValidNumber]));
+
+    //  convertTo :: TimeUnit -> ValidDate -> ValidNumber
+    var convertTo =
+    def('convertTo',
+        {},
+        [TimeUnit, $.ValidDate, $.ValidNumber],
+        function recur(unit, date) {
+          switch (unit) {
+            case 'milliseconds': return date.valueOf();
+            case 'seconds':      return recur('milliseconds', date) / 1000;
+            case 'minutes':      return recur('seconds', date) / 60;
+            case 'hours':        return recur('minutes', date) / 60;
+          }
+        });
+
+    throws(function() { convertTo('days', new Date(0)); },
+           errorEq(TypeError,
+                   '‘convertTo’ expected a value of type ' +
+                   '("milliseconds" | "seconds" | "minutes" | "hours") ' +
+                   'as its first argument; received "days"'));
+
+    eq(convertTo('seconds', new Date(1000)), 1);
+
+    //  SillyType :: Type
+    var SillyType = $.EnumType(['foo', true, 42]);
+
+    var _def = $.create($.env.concat([SillyType]));
+
+    //  id :: a -> a
+    var id = _def('id', {}, [a, a], R.identity);
+
+    eq(id(['foo', true]), ['foo', true]);
+
+    throws(function() { id(['foo', false]); },
+           errorEq(TypeError,
+                   '‘id’ received ["foo", false] as its first argument, ' +
+                   'but this value is not a member of any of the types ' +
+                   'in the environment:\n' +
+                   '\n' +
+                   '  - (Array ???)\n' +
+                   '  - Boolean\n' +
+                   '  - Date\n' +
+                   '  - Error\n' +
+                   '  - Function\n' +
+                   '  - Null\n' +
+                   '  - Number\n' +
+                   '  - Object\n' +
+                   '  - RegExp\n' +
+                   '  - String\n' +
+                   '  - Undefined\n' +
+                   '  - ("foo" | true | 42)'));
+  });
+
   it('supports record types', function() {
     //  Point :: Type
     var Point = $.RecordType({x: $.Number, y: $.Number});
@@ -552,7 +610,7 @@ describe('def', function() {
     var sinceEpoch = def('sinceEpoch',
                          {},
                          [$.ValidDate, $.Number],
-                         function(date) { return date.getTime() / 1000; });
+                         function(date) { return date.valueOf() / 1000; });
 
     throws(function() { sinceEpoch(new Date('foo')); },
            errorEq(TypeError,


davidchambers:
> Wasn't too sure about the String representation so I followed the example of `Nullable`.

Perhaps `("ms" | "seconds")` is better than `(Enum ("ms", "seconds"))`. This wouldn't provide a great way to represent an Enum with fewer than two possible values, but we shouldn't be bothered by such an absurd scenario. :)

svozza:
Yep, I prefer that.

README.md
```diff
@@ -162,6 +162,14 @@ $.Date :: Type
 
 Type comprising every Date value.
 
+#### `Enum`
+
+```haskell
+$.Enum :: [Any] -> Type
+```
+
+An enumerated type comprising a finite set of values.
```
svozza: This strikes me as tautologous but I'm not sure how to phrase it any better.

davidchambers: `$.Enum` is a type *constructor*. We can consult the description of `$.Array` for guidance:

> Constructor for homogeneous Array types.

Perhaps the following is sufficient?

> Constructor for [enumerated types][1].


[1]: https://en.wikipedia.org/wiki/Enumerated_type

index.js
```diff
@@ -356,6 +356,19 @@
     ($.Undefined  = type0('Undefined'))
   ];
 
+  // Enum :: [Any] -> Type
```
davidchambers: :arrow_right: :triangular_ruler:

index.js
```diff
@@ -356,6 +356,19 @@
...
+  $.Enum = function(constants) {
+    return {
+      type: 'sanctuary-def/Enum',
+      test: function(x) {
+        return $.Array(Any).test(constants) && constants.indexOf(x) !== -1;
```
davidchambers: What role does `$.Array(Any).test(constants)` serve? Verifying that `constants` is an Array? If so, I don't think this is appropriate. It's an error to apply `$.Enum` to a non-Array value. If we're to type check, we should do so up front. Since we don't do this elsewhere, though, I suggest not checking the type of `constants` at this point. If we decide to do so we should create a pull request to add type checking to all our type constructors.

README.md
```diff
@@ -162,6 +162,16 @@ $.Date :: Type
...
 
+#### `Enum`
+
+```haskell
+$.Enum :: [Any] -> Type
+```
+
+Constructor for [enumerated types][1].
+
+[1]: https://en.wikipedia.org/wiki/Enumerated_type
```
davidchambers: `[1]` is already defined in this document. You'll need to juggle the reference ids and slot this link into the list at the bottom of the document.

test/index.js
```diff
@@ -545,6 +545,28 @@ describe('def', function() {
                    'as its first argument; received ["abc"]'));
   });
 
+  it('supports "enum" types', function() {
+    var def = $.create($.env.concat([$.Enum, $.ValidDate, $.ValidNumber]));
+
+    //  convertTo :: Enum -> ValidDate -> ValidNumber
+    var convertTo =
+      def('convertTo',
```
davidchambers: Nit: Could you nudge this line (and the following six lines) two spaces to the left? That's the indentation we're using elsewhere.

index.js
svozza: Yeah, I was more just trying to ensure that the value I'm checking has an `indexOf` method. Should I just use:

```js
typeof constants.indexOf === 'function'
```

davidchambers: We should assume `constants` to be an Array. To support old versions of IE, though, we should avoid `Array.prototype.indexOf`. Also, it would be nice to use value-based equality. I suggest something like this:

```javascript
test: function(x) {
  var repr = show(x);
  var reprs = map(constants, show);
  for (var idx = 0; idx < reprs.length; idx += 1) {
    if (reprs[idx] === repr) return true;
  }
  return false;
}
```

Since `map(constants, show)` would then appear in both `test` and `toString`, we could even define `reprs` in the (innermost) scope shared by both functions.

test/index.js

README.md

index.js
```diff
@@ -356,6 +356,19 @@
...
+  //  Enum :: [Any] -> Type
+  $.Enum = function(constants) {
+    return {
+      type: 'sanctuary-def/Enum',
+      test: function(x) {
+        return $.Array(Any).test(constants) && constants.indexOf(x) !== -1;
+      },
+      toString: function() {
+        return '(' + map(constants, show).join(' | ') + ')';
+      }
```
davidchambers:
```diff
-      toString: function() {
-        return '(' + map(constants, show).join(' | ') + ')';
-      }
+      toString: always('(' + map(constants, show).join(' | ') + ')')
```

README.md
davidchambers: Thanks for making this fiddly change. :)

index.js
svozza: I don't think that last bit will work because `reprs` depends on `show(x)` and `x` is only available in `test`'s scope.

davidchambers: `reprs` is just a name for `map(constants, show)`, no?

svozza: Ah wait, I was confusing that with `_show`, which is a curried function of arity 2. Oops.

index.js

index.js
```diff
@@ -356,6 +356,22 @@
     ($.Undefined  = type0('Undefined'))
   ];
 
+  //  Enum :: [Any] -> Type
+  $.Enum = function(constants) {
+    var reprs = map(constants, show);
+    return {
+      type: 'sanctuary-def/Enum',
```
davidchambers: Let's change this value to `'ENUM'`, for consistency with `'NULLARY'`, `'UNARY'`, `'BINARY'`, `'RECORD'`, etc.

svozza: Yeah, I wasn't sure about that. Should I move the definition `Enum` up to where those types are? I only had it where it was because I was relying on that `$.Array` test.

test/index.js
```diff
@@ -545,6 +545,28 @@ describe('def', function() {
                    'as its first argument; received ["abc"]'));
   });
 
+  it('supports "enum" types', function() {
+    var def = $.create($.env.concat([$.Enum, $.ValidDate, $.ValidNumber]));
```
davidchambers: It seems to me we want this:

```javascript
//  TimeUnit :: Type
var TimeUnit = $.Enum(['ms', 'seconds']);

var def = $.create($.env.concat([TimeUnit, $.ValidDate, $.ValidNumber]));
```

As an aside, it bothers me *slightly* that we have `("ms" | "seconds")` rather than `("milliseconds" | "seconds")` or `("ms" | "s")`. ;)

svozza: I've prepended `TimeUnit` with a `$` to make it consistent with the convention you followed in plaid/sanctuary/pull/123

:zap:

README.md
```diff
@@ -162,6 +162,14 @@ $.Date :: Type
...
 Type comprising every Date value.
 
+#### `Enum`
+
+```haskell
+$.Enum :: [Any] -> Type
+```
+
+Constructor for [enumerated types][2].
+
```
davidchambers: Let's move this to the [type constructors][1] section—just before [`$.RecordType`][2], I suggest—and expand the documentation to match the format used for the other type constructors.


[1]: https://github.com/plaid/sanctuary-def/blob/master/README.md#type-constructors
[2]: https://github.com/plaid/sanctuary-def/blob/master/README.md#recordtype

davidchambers: Also, for consistency with `$.RecordType` we should name this `$.EnumType`. I realize that `$.Nullable` is the odd one out, but my thinking was that there's not much value in writing `const NullableString = $.NullableType($.String);` since one could simply write `$.NullableType($.String)` in situ. At that point it would be nice to be able to write `$.Nullable($.String)` instead, as it reads better. For enums and records, though, there's clearly an advantage to defining the type in just one place and reusing the definition.

test/index.js
```diff
@@ -545,6 +545,31 @@ describe('def', function() {
                    'as its first argument; received ["abc"]'));
   });
 
+  it('supports "enum" types', function() {
+    //  TimeUnit :: Type
+    var $TimeUnit = $.Enum(['milliseconds', 'seconds']);
```
davidchambers: The identifier does not match the type signature. ;)

test/index.js
davidchambers: I'm happy with either `TimeUnit` or `$TimeUnit`. In Sanctuary my hand was forced, as I wanted to define Maybe and Either types but the `Maybe` and `Either` identifiers were already taken. :)

svozza: Ah right, I actually prefer it without the `$` so I've removed them.

README.md
svozza: Yep, makes sense.

:zap:

test/index.js

README.md
```diff
@@ -542,6 +542,34 @@ showCard(Pair('X', '♠'));
 // ! TypeError: ‘showCard’ expected a value of type (Pair Rank Suit) as its first argument; received Pair("X", "♠")
 ```
 
+#### `EnumType`
+
+`EnumType` is used to construct [enumerated types][5].
+
+```javascript
+//    TimeUnit :: Type
+const TimeUnit = $.Enum(['milliseconds', 'seconds']);
```
davidchambers: `s/Enum/EnumType/`

svozza: Oops. :zap:

README.md
```diff
@@ -542,6 +542,34 @@ showCard(Pair('X', '♠'));
...
+convertTo('milliseconds', new Date(1));
+// => 1000
+
+convertTo('days', new Date(1));
+// => ! TypeError: ‘convertTo’ expected a value of type ("milliseconds" | "seconds") as its first argument; received "days"
+```
+
+```haskell
+$.EnumType :: [Any] -> Type
+```
```
davidchambers: This should precede the example, as with the `$.RecordType` documentation. Also, we're a bit inconsistent: we don't include the `$.` prefix in the type constructor signatures. To be consistently inconsistent, let's drop the prefix. :)

index.js
```diff
@@ -280,6 +280,22 @@
     });
   };
 
+  //  Enum :: [Any] -> Type
```
davidchambers: `s/Enum/EnumType/`

README.md
svozza: Hehe. Understood.

:zap

index.js

index.js
```diff
@@ -280,6 +280,22 @@
     });
   };
 
+  //  EnumType :: [Any] -> Type
+  $.EnumType = function(constants) {
+    var reprs = map(constants, show);
```
davidchambers: > In computer programming, an __enumerated type__ […] is a data type consisting of a set of named values called __elements__, __members__, __enumeral__, or __enumerators__ of the type.

I'd like to use one of these words in place of `constants`. I think of lists as having elements and sets as having members, so `members` gets my vote. :)

test/index.js
```diff
@@ -464,6 +464,31 @@ describe('def', function() {
                    'as its first argument; received 0.5'));
   });
 
+  it('supports "enum" types', function() {
```
davidchambers: Nit: I'd use "enumerated" here (unquoted). We needn't save characters in this context. :)

index.js
svozza: Yes, `constants` didn't sit quite right with me too.

test/index.js

README.md
```diff
@@ -542,6 +542,36 @@ showCard(Pair('X', '♠'));
...
+def('convertTo',
+    {},
+    [TimeUnit, $.ValidDate, $.ValidNumber],
+    (unit, date) => date.valueOf() * (unit === 'milliseconds' ? 1000 : 1));
+    
+convertTo('milliseconds', new Date(1));
+// => 1000
+
+convertTo('days', new Date(1));
+// => ! TypeError: ‘convertTo’ expected a value of type ("milliseconds" | "seconds") as its first argument; received "days"
```
davidchambers: The `=>` is intended to be read _evaluates to_. The `!` is intended to be read _throws_. We want `!` here but not `=>`.

test/index.js
```diff
@@ -464,6 +464,31 @@ describe('def', function() {
...
+    var def = $.create($.env.concat([TimeUnit, $.ValidDate, $.ValidNumber]));
+
+    //  convertTo :: TimeUnit -> ValidDate -> ValidNumber
+    var convertTo =
+    def('convertTo',
+        {},
+        [TimeUnit, $.ValidDate, $.ValidNumber],
+        function(unit, date) {
+          var multiplier = unit === 'milliseconds' ? 1000 : 1;
+          return date.valueOf() * multiplier;
```
davidchambers: Your `sinceEpoch` function uses `.getTime()`. Feel free to do the same here if you prefer it.

svozza: Seen as though we use `valueOf` in the docs and elsewhere in the code, I'll change `sinceEpoch`.

README.md
```diff
@@ -542,6 +542,36 @@ showCard(Pair('X', '♠'));
 // ! TypeError: ‘showCard’ expected a value of type (Pair Rank Suit) as its first argument; received Pair("X", "♠")
 ```
 
+#### `EnumType`
+
+`EnumType` is used to construct [enumerated types][5].
+
+```haskell
```
davidchambers: We're missing one paragraph present in the `$.RecordType` documentation, along the lines of the following:

```markdown
To define an enumerated type one must provide:

  - an array of values with distinct [`R.toString`][6] representations.
```

test/index.js
davidchambers: Okay. I'm primarily concerned with internal consistency within the source code, so this is one of the *rare* occasions when I'm not fussed. :)

README.md

test/index.js
svozza: In for a penny, in for a pound!

davidchambers:
We should update `equalTypes` to recognize enumerated types. I believe duplicating the `'RECORD'` case and changing the value to `'ENUM'` should do the trick.

svozza:
Good catch, I would have completely missed that one.

svozza:
Hmm, how do the other parts of that `case` statement get tested?

davidchambers:
> Hmm, how do the other parts of that `case` statement get tested?

Let me investigate.

davidchambers:
This'll restore coverage to 100%:

```javascript
//  SillyType :: Type
var SillyType = $.EnumType(['foo', true, 42]);

var def = $.create($.env.concat([SillyType]));

//  id :: a -> a
var id = def('id', {}, [a, a], R.identity);

eq(id(['foo', true]), ['foo', true]);

throws(function() { id(['foo', false]); },
       errorEq(TypeError,
               '‘id’ received ["foo", false] as its first argument, ' +
               'but this value is not a member of any of the types ' +
               'in the environment:\n' +
               '\n' +
               '  - (Array ???)\n' +
               '  - Boolean\n' +
               '  - Date\n' +
               '  - Error\n' +
               '  - Function\n' +
               '  - Null\n' +
               '  - Number\n' +
               '  - Object\n' +
               '  - RegExp\n' +
               '  - String\n' +
               '  - Undefined\n' +
               '  - ("foo" | true | 42)'));
```

svozza:
Nice one! I wasn't sure where to put it so I just added it in with the other tests. I'm wondering should I separate the two tests by adding ```it("can define a custom enumerated type)`` and ```it("can be tested for equality)``` functions.

README.md
```diff
@@ -542,6 +542,40 @@ showCard(Pair('X', '♠'));
...
+const TimeUnit = $.EnumType(['milliseconds', 'seconds']);
+
+const def = $.create($.env.concat([TimeUnit, $.ValidDate, $.ValidNumber]));
+
+//    convertTo :: TimeUnit -> ValidDate -> ValidNumber
+const convertTo =
+def('convertTo',
+    {},
+    [TimeUnit, $.ValidDate, $.ValidNumber],
+    (unit, date) => date.valueOf() * (unit === 'milliseconds' ? 1000 : 1));
```
davidchambers: There's a logic error here, I believe: `date.valueOf()` returns the time in milliseconds—rather than seconds—since the epoch, so we should divide by 1000 in the `'seconds'` case rather multiply by 1000 in the `'milliseconds'` case.

We could use `switch` (and increase the number of members) to make `TimeUnit`'s enumerated nature more apparent:

```javascript
function recur(unit, date) {
  switch (unit) {
    'milliseconds': return date.valueOf();
    'seconds':      return recur('milliseconds', date) / 1000;
    'minutes':      return recur('seconds', date) / 60;
    'hours':        return recur('minutes', date) / 60;
  }
}
```

svozza: Of course, I don't know why I thought that function returned seconds rather than milliseconds.

test/index.js
```diff
@@ -464,6 +464,60 @@ describe('def', function() {
...
+
+    //  convertTo :: TimeUnit -> ValidDate -> ValidNumber
+    var convertTo =
+    def('convertTo',
+        {},
+        [TimeUnit, $.ValidDate, $.ValidNumber],
+        function(unit, date) {
+          var divisor = unit === 'seconds' ? 1000 : 1;
+          return date.valueOf() / divisor;
+        });
```
davidchambers: Let's update this to match the example in the README.

test/index.js
```diff
@@ -464,6 +464,64 @@ describe('def', function() {
                    'as its first argument; received 0.5'));
   });
 
+  it('supports enumerated types', function() {
+    //  TimeUnit :: Type
+    var TimeUnit = $.EnumType(['milliseconds', 'seconds']);
```
davidchambers: We should add `'minutes'` and `'hours'` as members.

README.md
```diff
@@ -542,6 +542,48 @@ showCard(Pair('X', '♠'));
...
+        case 'hours':        return recur('minutes', date) / 60;
+      }
+    });
+    
+convertTo('seconds', new Date(1000));
+// => 1
+
+convertTo('days', new Date(1));
+// ! TypeError: ‘convertTo’ expected a value of type ("milliseconds" | "seconds" | "minutes" | "hours") 
+     as its first argument; received "days"
```
davidchambers: Let's not wrap this message, as it looks wrong when rendered.

README.md
```diff
@@ -542,6 +542,48 @@ showCard(Pair('X', '♠'));
...
+        case 'seconds':      return recur('milliseconds', date) / 1000;
+        case 'minutes':      return recur('seconds', date) / 60;
+        case 'hours':        return recur('minutes', date) / 60;
+      }
+    });
+    
+convertTo('seconds', new Date(1000));
+// => 1
+
+convertTo('days', new Date(1));
```
davidchambers: Let's use `new Date(1000)` as the second argument here as well, to draw the reader's attention to the first argument.

README.md

README.md

test/index.js
```diff
@@ -464,6 +464,64 @@ describe('def', function() {
...
+            case 'minutes':      return recur('seconds', date) / 60;
+            case 'hours':        return recur('minutes', date) / 60;
+          }
+        });
+
+    throws(function() { convertTo('days', new Date(0)); },
+           errorEq(TypeError,
+                   '‘convertTo’ expected a value of type ' +
+                   '("milliseconds" | "seconds" | "minutes" | "hours") as ' +
+                   'its first argument; received "days"'));
```
davidchambers: I feel bad even requesting this change, as it's the nittiest of nits. I like to wrap these messages after the type representation, if possible. In this case:

```diff
                    '‘convertTo’ expected a value of type ' +
-                   '("milliseconds" | "seconds" | "minutes" | "hours") as ' +
-                   'its first argument; received "days"'));
+                   '("milliseconds" | "seconds" | "minutes" | "hours") ' +
+                   'as its first argument; received "days"'));
```

Thank you for accommodating my fastidiousness!

README.md
```diff
@@ -542,6 +542,47 @@ showCard(Pair('X', '♠'));
...
+        case 'minutes':      return recur('seconds', date) / 60;
+        case 'hours':        return recur('minutes', date) / 60;
+      }
+    });
+    
+convertTo('seconds', new Date(1000));
+// => 1
+
+convertTo('days', new Date(1000));
+// ! TypeError: ‘convertTo’ expected a value of type ("milliseconds" | "seconds" | "minutes" | "hours")  as its first argument; received "days"
```
davidchambers: There appear to be two spaces after the `)`.

test/index.js
svozza: Not at all, a lot of these were silly mistakes I should have caught myself. I'm just going to blame tiredness and thus absolve myself of all culpability. ;)

davidchambers:
:deciduous_tree:

test/index.js
davidchambers: :waxing_crescent_moon:

PR #18, opened 2016-01-09
Add Number Types

svozza:
As discussed in #11, I've added the Number types. I'll start on the Integer types next but I'll wait until #13 is merged before rebasing this PR and starting the new one.

PR #21, opened 2016-01-13
Add integer types

svozza:
As per #11 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ $.FiniteNumber :: Type
 ```
 
 Type comprising every [`ValidNumber`](#validnumber) value except `Infinity` and 
-`-Infinity` (and their object counterparts). 
+`-Infinity` (and their object counterparts).
 
 #### `Function`
 
@@ -188,6 +188,23 @@ $.Function :: Type
 
 Type comprising every Function value.
 
+#### `Integer`
+
+```haskell
+$.Integer :: Type
+```
+
+Type comprising every integer in the range 
+[[`Number.MIN_SAFE_INTEGER`][4] .. [`Number.MAX_SAFE_INTEGER`][5]].
+
+#### `NegativeInteger`
+
+```haskell
+$.NegativeInteger :: Type
+```
+
+Type comprising every [`Integer`](#integer) value less than zero.
+
 #### `NonZeroFiniteNumber`
 
 ```haskell
@@ -195,7 +212,15 @@ $.NonZeroFiniteNumber :: Type
 ```
 
 Type comprising every [`FiniteNumber`](#finitenumber) value except `0` and `-0` 
-(and their object counterparts). 
+(and their object counterparts).
+
+#### `NonZeroInteger`
+
+```haskell
+$.NonZeroInteger :: Type
+```
+
+Type comprising every non-zero [`Integer`](#integer) value.
 
 #### `NonZeroValidNumber`
 
@@ -204,7 +229,7 @@ $.NonZeroValidNumber :: Type
 ```
 
 Type comprising every [`ValidNumber`](#validnumber) value except `0` and `-0`
-(and their object counterparts). 
+(and their object counterparts).
 
 #### `Null`
 
@@ -239,9 +264,17 @@ $.Object :: Type
 Type comprising every "plain" Object value. Specifically, values created via:
 
   - object literal syntax;
-  - [`Object.create`][4]; or
+  - [`Object.create`][6]; or
   - the `new` operator in conjunction with `Object` or a custom
     constructor function.
+
+#### `PositiveInteger`
+
+```haskell
+$.PositiveInteger :: Type
+```
+
+Type comprising every [`Integer`](#integer) value greater than zero.
 
 #### `RegExp`
 
@@ -552,11 +585,11 @@ showCard(Pair('X', '♠'));
 
 #### `EnumType`
 
-`EnumType` is used to construct [enumerated types][5].
+`EnumType` is used to construct [enumerated types][7].
 
 To define an enumerated type one must provide:
 
-  - an array of values with distinct [`R.toString`][6] representations.
+  - an array of values with distinct [`R.toString`][8] representations.
 
 ```haskell
 EnumType :: [Any] -> Type
@@ -659,8 +692,8 @@ _concat([1, 2], 'buzz');
 
 The type of `_concat` is misleading: it suggests that it can operate on any
 two values of *any* one type. In fact there's an implicit constraint, since
-the type must support concatenation (in [mathematical][7] terms, the type
-must have a [semigroup][8]). The run-time type errors that result when this
+the type must support concatenation (in [mathematical][9] terms, the type
+must have a [semigroup][10]). The run-time type errors that result when this
 constraint is violated are not particularly descriptive:
 
 ```javascript
@@ -702,8 +735,10 @@ Multiple constraints may be placed on a type variable by including multiple
 [1]: http://ramdajs.com/docs/#__
 [2]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SyntaxError
 [3]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError
-[4]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create
-[5]: https://en.wikipedia.org/wiki/Enumerated_type
-[6]: http://ramdajs.com/docs/#toString
-[7]: https://en.wikipedia.org/wiki/Semigroup
-[8]: https://github.com/fantasyland/fantasy-land#semigroup
+[4]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MIN_SAFE_INTEGER
+[5]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER
+[6]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create
+[7]: https://en.wikipedia.org/wiki/Enumerated_type
+[8]: http://ramdajs.com/docs/#toString
+[9]: https://en.wikipedia.org/wiki/Semigroup
+[10]: https://github.com/fantasyland/fantasy-land#semigroup
--- a/index.js
+++ b/index.js
@@ -19,6 +19,9 @@
 
   var $ = {};
 
+  var MAX_SAFE_INTEGER = Math.pow(2, 53) - 1;
+  var MIN_SAFE_INTEGER = -MAX_SAFE_INTEGER;
+
   var LEFT_SINGLE_QUOTATION_MARK = '\u2018';
   var RIGHT_SINGLE_QUOTATION_MARK = '\u2019';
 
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -188,6 +188,15 @@ $.Function :: Type
...
 Type comprising every Function value.
 
+#### `Integer`
+
+```haskell
+$.Integer :: Type
+```
+
+Type comprising every non-floating point [`ValidNumber`](#validnumber) value 
+between the range `-9007199254740992` and `9007199254740992`.
```
davidchambers: How about this?

> Type comprising every integer in the range [[`Number.MIN_SAFE_INTEGER`][1] .. [`Number.MAX_SAFE_INTEGER`][2]].


[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MIN_SAFE_INTEGER
[2]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

README.md
```diff
@@ -197,6 +206,15 @@ $.NonZeroFiniteNumber :: Type
...
 (and their object counterparts). 
 
+#### `NonZeroInteger`
+
+```haskell
+$.NonZeroInteger :: Type
+```
+
+Type comprising every [`Integer`](#integer) value except `0` and `-0` (and 
+their object counterparts).
```
davidchambers: I think this would suffice:

> Type comprising every non-zero [`Integer`](#integer) value.

README.md
svozza: I'll wait until #20 lands on master before that because I don't want to have to do the references at the bottom of the README twice.

README.md

README.md
```diff
@@ -235,6 +253,22 @@ Type comprising every "plain" Object value. Specifically, values created via:
...
+
+Type comprising every [`Integer`](#integer) value greater than zero.
+
+#### `NegativeInteger`
+
+```haskell
+$.NegativeInteger :: Type
+```
+
+Type comprising every [`Integer`](#integer) value less than zero.
```
davidchambers: This is a clear, succinct description. :)

I do wonder, though, whether we should obey strict alphabetical ordering. If we're to group *some* related types it seems to me we should group *all* related types. Grouping has the advantage of making the types section more cohesive upon first reading, but it makes it harder to find a section by name. It also forces us to decide to which group a type belongs each time we add one. Strict alphabetical ordering seems right for now. If we decide to group the types we should discuss this as a separate change.

README.md
davidchambers: I was thinking the same thing. :)

README.md
svozza: Yep, I agree. That was just a mistake on my part. The list was meant to be in alphabetical order.

README.md
```diff
@@ -188,6 +188,23 @@ $.Function :: Type
 
 Type comprising every Function value.
 
+#### `Integer`
+
+```haskell
+$.Integer :: Type
+```
+
+Type comprising every integer in the range [[`Number.MIN_SAFE_INTEGER`][4] .. [`Number.MAX_SAFE_INTEGER`][5]].
```
davidchambers: Let's wrap this line.

README.md
```diff
@@ -188,6 +188,23 @@ $.Function :: Type
...
+
+
```
davidchambers: :space_invader: We only need one empty line here.

README.md
```diff
@@ -204,7 +229,7 @@ $.NonZeroValidNumber :: Type
 ```
 
 Type comprising every [`ValidNumber`](#validnumber) value except `0` and `-0`
-(and their object counterparts). 
+(and their object counterparts).
```
davidchambers: There are a few more occurrences of trailing whitespace in this file. Mind trimming these while you're at it?

index.js
```diff
@@ -399,6 +399,36 @@
     function(x) { return $.FiniteNumber.test(x) && Number(x) !== 0; }
   );
 
+  //  Integer :: Type
+  $.Integer = $.NullaryType(
+    'sanctuary-def/Integer',
+    function(x) {
+      var max = Math.pow(2, 53) - 1;
```
davidchambers: I'd like to define `MAX_SAFE_INTEGER` and `MIN_SAFE_INTEGER` "constants" as in [Sanctuary][1]. These definitions should live near the top of the file:

```diff
   'use strict';

   var $ = {};

+  var MAX_SAFE_INTEGER = Math.pow(2, 53) - 1;
+  var MIN_SAFE_INTEGER = -MAX_SAFE_INTEGER;
+
   var LEFT_SINGLE_QUOTATION_MARK = '\u2018';
   var RIGHT_SINGLE_QUOTATION_MARK = '\u2019';
```


[1]: https://github.com/plaid/sanctuary/blob/v0.7.1/index.js#L121-L122

index.js
```diff
@@ -399,6 +399,36 @@
...
+  //  NonZeroInteger :: Type
+  $.NonZeroInteger = $.NullaryType(
+    'sanctuary-def/NonZeroInteger',
+    function(x) { return $.NonZeroValidNumber.test(x) && $.Integer.test(x); }
+  );
+
+  //  PositiveInteger :: Type
+  $.PositiveInteger = $.NullaryType(
+    'sanctuary-def/PositiveInteger',
+    function(x) { return $.NonZeroInteger.test(x) && x > 0; }
```
davidchambers: Let's use `$.Integer.test` rather than `$.NonZeroInteger.test` here. `x > 0` handles zeros.

index.js
```diff
@@ -399,6 +399,36 @@
...
+  );
+
+  //  NegativeInteger :: Type
+  $.NegativeInteger = $.NullaryType(
+    'sanctuary-def/NegativeInteger',
+    function(x) { return $.NonZeroInteger.test(x) && x < 0; }
```
davidchambers: Let's use `$.Integer.test` instead.

README.md

README.md

README.md

index.js

index.js

index.js

index.js
```diff
@@ -399,6 +402,35 @@
     function(x) { return $.FiniteNumber.test(x) && Number(x) !== 0; }
   );
 
+  //  Integer :: Type
+  $.Integer = $.NullaryType(
+    'sanctuary-def/Integer',
+    function(x) {
+      return $.ValidNumber.test(x) &&
+        Math.floor(x) === Number(x) &&
+        x >= MIN_SAFE_INTEGER  &&
```
davidchambers: Nit: There appear to be two spaces before the `&&` operator.

index.js
```diff
@@ -399,6 +402,35 @@
...
+        x <= MAX_SAFE_INTEGER;
+    }
+  );
+
+  //  NonZeroInteger :: Type
+  $.NonZeroInteger = $.NullaryType(
+    'sanctuary-def/NonZeroInteger',
+    function(x) { return $.NonZeroValidNumber.test(x) && $.Integer.test(x); }
```
davidchambers: I'd prefer to let `$.Integer.test` reject `Infinity`, `-Infinity`, and `NaN`. We could then have a nice symmetry between this predicate and the following two:

```javascript
//  $.NonZeroInteger.test
x => $.Integer.test(x) && x != 0

//  $.PositiveInteger.test
x => $.Integer.test(x) && x > 0

//  $.NegativeInteger.test
x => $.Integer.test(x) && x < 0
```

test/index.js
```diff
@@ -766,6 +766,136 @@ describe('def', function() {
...
+           errorEq(TypeError,
+                   '‘inc’ expected a value of type Integer ' +
+                   'as its first argument; received ' +
+                   'new Number(-9007199254740992)'));
+
+    throws(function() { inc(new Number(-9007199254740992)); },
+           errorEq(TypeError,
+                   '‘inc’ expected a value of type Integer ' +
+                   'as its first argument; received ' +
+                   'new Number(-9007199254740992)'));
```
davidchambers: The two tests above are identical.

test/index.js
```diff
@@ -766,6 +766,136 @@ describe('def', function() {
...
+  it('supports the "PositiveInteger" type', function() {
+    var def = $.create($.env.concat([$.PositiveInteger]));
+
+    //  inc :: PositiveInteger -> PositiveInteger
+    var inc = def('inc',
+                  {},
+                  [$.PositiveInteger, $.PositiveInteger],
+                  function(x) { return x + 1; });
+
+    throws(function() { inc(-1.5); },
```
davidchambers: `-1.5` is neither positive nor an integer. Since all the invalid input values are currently negative, we're not necessarily testing the "integer" check. We could use `1.5` here instead.

test/index.js
```diff
@@ -766,6 +766,136 @@ describe('def', function() {
...
+  it('supports the "NegativeInteger" type', function() {
+    var def = $.create($.env.concat([$.NegativeInteger]));
+
+    //  dec :: NegativeInteger -> NegativeInteger
+    var dec = def('dec',
+                  {},
+                  [$.NegativeInteger, $.NegativeInteger],
+                  function(x) { return x - 1; });
+
+    throws(function() { dec(1.5); },
```
davidchambers: Same as above. We should use `-1.5` here instead to ensure negative non-integer values are rejected.

index.js
svozza: I'd love to do that but as noted in #18, the linter won't allow `!=`

index.js
```diff
@@ -399,6 +402,35 @@
...
 
+  //  Integer :: Type
+  $.Integer = $.NullaryType(
+    'sanctuary-def/Integer',
+    function(x) {
+      return $.ValidNumber.test(x) &&
+        Math.floor(x) === Number(x) &&
+        x >= MIN_SAFE_INTEGER  &&
+        x <= MAX_SAFE_INTEGER;
+    }
```
davidchambers: I like to align the subexpressions when wrapping after `&&`:

```javascript
function(x) {
  return $.ValidNumber.test(x) &&
         Math.floor(x) === Number(x) &&
         x >= MIN_SAFE_INTEGER &&
         x <= MAX_SAFE_INTEGER;
}
```

Also, should we replace `Math.floor(x) === Number(x)` with `Math.floor(x) == x` do you think?

index.js
davidchambers: Let's locate that rule and nix it. ;)

index.js
svozza: Yeah, that's how I had it originally and then I saw in the `Integer` definition in `test\index.js` that the object constructor was used so I changed it.

test/index.js

test/index.js

test/index.js

index.js
svozza: So I just disabled the relevant checks for those specific lines that use the `!=` operator.

index.js

index.js
```diff
@@ -387,17 +390,48 @@
     function(x) { return $.ValidNumber.test(x) && isFinite(x); }
   );
 
+  //  Integer :: Type
+  $.Integer = $.NullaryType(
+    'sanctuary-def/Integer',
+    function(x) {
+      return $.ValidNumber.test(x) &&
+             Math.floor(x) === x &&
```
davidchambers: This needs to be `==`, doesn't it?

```javascript
> Math.floor(new Number(42)) === new Number(42)
false
```

index.js
```diff
@@ -387,17 +390,48 @@
...
+             x >= MIN_SAFE_INTEGER &&
+             x <= MAX_SAFE_INTEGER;
+    }
+  );
+
+  /*jshint -W041, -W116*/
```
davidchambers: I'm happy to remove these rules entirely. I have faith in our ability to flag inappropriate use of `==`/`!=` during reviews.

index.js
svozza: Oops. :zap:

index.js
svozza: Agreed. :zap:

index.js
```diff
@@ -390,13 +393,42 @@
...
+  );
+
+  //  Integer :: Type
+  $.Integer = $.NullaryType(
+    'sanctuary-def/Integer',
+    function(x) {
+      return $.ValidNumber.test(x) &&
+        Math.floor(x) == x &&
+        x >= MIN_SAFE_INTEGER &&
+        x <= MAX_SAFE_INTEGER;
```
davidchambers: Mind indenting these three lines to align with `$`?

svozza: Autoformatting... :angry:

davidchambers:
:deciduous_tree:

svozza:
:tada:

PR #22, opened 2016-01-14
Run tests against Node 4

svozza:
Node 4 is the officially released version of node.js with LTS. We should be running our tests against it. I've also added tests for Node 5. Luckily these are already installed on the CircleCI container so we can just use them with `nvm use`. If we were to drop support for Node 0.10, which is EOL in October, then we could set the base Node version to 0.12.7 and not have to do any installs at all.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -13,3 +13,5 @@ test:
     - nvm install 0.12.7
     - nvm alias default 0.12.7
     - make test
+    - nvm use 4 && make test
+    - nvm use 5 && make test


circle.yml
```diff
@@ -13,3 +13,5 @@ test:
     - nvm install 0.12.7
     - nvm alias default 0.12.7
     - make test
+    - nvm use 4.2 && make test
+    - nvm use 5 && make test
```
davidchambers: What's the thinking behind specifying `(4.2, 5)` rather than `(4, 5)` or `(4.2, 5.4)`?

svozza: I used 4.2 because it's the official LTS version. Node 5.2 (or any version of 5) won't be eligible for LTS and it's not installed by default on CircleCI so I didn't want to slow down the build by doing another install and figured using 5.0, which is installed by default, would be okay.

davidchambers: That makes sense. Is `5` equivalent to `5.0` or does it mean _the latest 5.x.x version_?

svozza:
Actually, you're right `nvm 5` installs the latest version of Node so I should use `4` and `5` there.

svozza:
Actually wait, I'm wrong there, `nvm 5` will use the latest installed version of `node` but when I did `nvm use 5.4` I got an error saying it wasn't installed. So what's happening there is that we're running against the latest 5.x version that happens to be in the container.

davidchambers:
:deciduous_tree: Merge this if you're happy with the current state of the pull request. :)

PR #24, opened 2016-01-18
types: define {Positive,Negative}{Number,FiniteNumber}

davidchambers:
This pull request introduces the following types:

  - `PositiveNumber`
  - `NegativeNumber`
  - `PositiveFiniteNumber`
  - `NegativeFiniteNumber`

`PositiveValidNumber` would be redundant, since `PositiveNumber` excludes `NaN` by virtue or `NaN` not being positive. Similar logic applies to `NegativeValidNumber`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -373,62 +373,86 @@
   ];
 
   //  ValidDate :: Type
-  $.ValidDate = $.NullaryType(
+  $.ValidDate = NullaryType(
     'sanctuary-def/ValidDate',
     function(x) { return $.Date.test(x) && !isNaN(x.valueOf()); }
   );
 
+  //  PositiveNumber :: Type
+  $.PositiveNumber = NullaryType(
+    'sanctuary-def/PositiveNumber',
+    function(x) { return $.Number.test(x) && x > 0; }
+  );
+
+  //  NegativeNumber :: Type
+  $.NegativeNumber = NullaryType(
+    'sanctuary-def/NegativeNumber',
+    function(x) { return $.Number.test(x) && x < 0; }
+  );
+
   //  ValidNumber :: Type
-  $.ValidNumber = $.NullaryType(
+  var ValidNumber = $.ValidNumber = NullaryType(
     'sanctuary-def/ValidNumber',
     function(x) { return $.Number.test(x) && !isNaN(x); }
   );
 
-  //  FiniteNumber :: Type
-  $.FiniteNumber = $.NullaryType(
-    'sanctuary-def/FiniteNumber',
-    function(x) { return $.ValidNumber.test(x) && isFinite(x); }
+  //  NonZeroValidNumber :: Type
+  $.NonZeroValidNumber = NullaryType(
+    'sanctuary-def/NonZeroValidNumber',
+    function(x) { return ValidNumber.test(x) && x != 0; }
   );
 
-  //  NonZeroValidNumber :: Type
-  $.NonZeroValidNumber = $.NullaryType(
-    'sanctuary-def/NonZeroValidNumber',
-    function(x) { return $.ValidNumber.test(x) && x != 0; }
+  //  FiniteNumber :: Type
+  var FiniteNumber = $.FiniteNumber = NullaryType(
+    'sanctuary-def/FiniteNumber',
+    function(x) { return ValidNumber.test(x) && isFinite(x); }
+  );
+
+  //  PositiveFiniteNumber :: Type
+  $.PositiveFiniteNumber = NullaryType(
+    'sanctuary-def/PositiveFiniteNumber',
+    function(x) { return FiniteNumber.test(x) && x > 0; }
+  );
+
+  //  NegativeFiniteNumber :: Type
+  $.NegativeFiniteNumber = NullaryType(
+    'sanctuary-def/NegativeFiniteNumber',
+    function(x) { return FiniteNumber.test(x) && x < 0; }
   );
 
   //  NonZeroFiniteNumber :: Type
-  $.NonZeroFiniteNumber = $.NullaryType(
+  $.NonZeroFiniteNumber = NullaryType(
     'sanctuary-def/NonZeroFiniteNumber',
-    function(x) { return $.FiniteNumber.test(x) && x != 0; }
+    function(x) { return FiniteNumber.test(x) && x != 0; }
   );
 
   //  Integer :: Type
-  $.Integer = $.NullaryType(
+  var Integer = $.Integer = NullaryType(
     'sanctuary-def/Integer',
     function(x) {
-      return $.ValidNumber.test(x) &&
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -373,62 +373,86 @@
...
-  $.ValidDate = $.NullaryType(
+  $.ValidDate = NullaryType(
     'sanctuary-def/ValidDate',
     function(x) { return $.Date.test(x) && !isNaN(x.valueOf()); }
   );
 
+  //  PositiveNumber :: Type
+  $.PositiveNumber = NullaryType(
+    'sanctuary-def/PositiveNumber',
+    function(x) { return $.Number.test(x) && x > 0; }
```
wmrowan: You declare `ValidNumber` below as `var ValidNumber = $ValidNumber =` so that you may use `ValidNumber` for other type definitions elsewhere in this scope. Here though you refer to `$.Number` because the way the built in types are declared. Why the inconsistency? And, for that matter, why are these types not included in `$.env`?

index.js
```diff
@@ -373,62 +373,86 @@
...
+
   //  ValidNumber :: Type
-  $.ValidNumber = $.NullaryType(
+  var ValidNumber = $.ValidNumber = NullaryType(
     'sanctuary-def/ValidNumber',
     function(x) { return $.Number.test(x) && !isNaN(x); }
   );
 
+  //  NonZeroValidNumber :: Type
+  $.NonZeroValidNumber = NullaryType(
```
wmrowan: Do you think we need a `NonZeroNumber` that includes `NaN` for completeness?

index.js
davidchambers: > Why the inconsistency?

Because I don't want to shadow [`Number`][1]. I completely agree that this is inconsistent, but I don't currently see an elegant alternative.

> And, for that matter, why are these types not included in `$.env`?

The more types the environment contains, the more types may be listed in error messages:

```javascript
$.create($.env)('f', {}, [a, a], x => 'abc')(42);
// ! TypeError: ‘f’ is expected to return a value of type Number; returned "abc"

$.create($.env.concat([$.PositiveNumber, $.ValidNumber, $.NonZeroValidNumber, $.FiniteNumber, $.PositiveFiniteNumber, $.NonZeroFiniteNumber, $.Integer, $.PositiveInteger, $.NonZeroInteger]))('f', {}, [a, a], x => 'abc')(42);
// ! TypeError: ‘f’ is expected to return a value of type Number or PositiveNumber or ValidNumber or NonZeroValidNumber or FiniteNumber or PositiveFiniteNumber or NonZeroFiniteNumber or Integer or PositiveInteger or NonZeroInteger; returned "abc"
```

Since the number of types we could define is infinite, we must decide which types "deserve" to be included in `$.env`. Restricting `$.env` to built-ins provides clear delineation.


[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number

index.js
davidchambers: Good question. I see the appeal, but I wonder whether it might only be used in situations where the author *intended* to use `NonZeroValidNumber`. In other words, would we be setting a trap by giving such a convenient name to such an "unlikely" type?

wmrowan: You could make a similar argument about `PositiveNumber` and `PositiveFiniteNumber`. I might guess that much of the time an author uses `PositiveNumber` he or she really didn't mean to included `Infinity`. But yes, I get your point.

index.js
wmrowan: The shadowing point makes sense.

> The more types the environment contains, the more types may be listed in error messages

This is going to be a problem whenever environments with large hierarchies are used. Is your expectation that users will be restrictive in the types they use in their environments? I bet many applications could get away with excluding `$.Number` and including only `$.PositiveInteger`.

Do you think we might include other, pre-built, environments that provide other convenient groupings?

index.js
davidchambers: > You could make a similar argument about `PositiveNumber` and `PositiveFiniteNumber`.

You're quite right. I can, though, imagine using `PositiveNumber` whereas I can't imagine using `NonZeroNumber`. Let's omit the latter for now. We're free to change our minds and add it later.

wmrowan:
:earth_americas:

index.js
davidchambers: > Do you think we might include other, pre-built, environments that provide other convenient groupings?

We could do, though I'm leaning in the other direction. We can't know which types will be relevant in a given context, so there's an argument to be made for requiring types to be selected à la carte. The main role of `$.env` is to make it easy to "set up" the library in the REPL, but `[$.Array, $.Boolean, $.Number, $.String]` would likely perform this role equally well.

Were we to provide pre-built environments, what might we name them and which types might they comprise?

test/index.js
```diff
@@ -648,6 +648,40 @@ describe('def', function() {
...
+    eq($.PositiveNumber.test(new Number(0)), false);
+    eq($.PositiveNumber.test(-0), false);
+    eq($.PositiveNumber.test(new Number(-0)), false);
+    eq($.PositiveNumber.test(0.5), true);
+    eq($.PositiveNumber.test(new Number(0.5)), true);
+    eq($.PositiveNumber.test(Infinity), true);
+    eq($.PositiveNumber.test(new Number(Infinity)), true);
+  });
+
+  it('supports the "NegativeNumber" type', function() {
```
svozza: I like this, it's a much nicer way of testing these types than what I've been doing.

index.js
svozza: I think we have to draw the line somewhere, I'm not sure there's much value in a `NonZeroNumber` type.

test/index.js
```diff
@@ -648,6 +648,40 @@ describe('def', function() {
...
+    eq($.NegativeNumber.name, 'sanctuary-def/NegativeNumber');
+    eq($.NegativeNumber.test(null), false);
+    eq($.NegativeNumber.test(NaN), false);
+    eq($.NegativeNumber.test(new Number(NaN)), false);
+    eq($.NegativeNumber.test(1), false);
+    eq($.NegativeNumber.test(new Number(1)), false);
+    eq($.NegativeNumber.test(0), false);
```
svozza: I know you prefer to keep dependencies within a test to a minimum but this pattern of checking the primitive and boxed versions of a type is so common I wonder should we have a convenience function for it.

davidchambers: It's certainly a possibility. Open an issue if you think of an elegant approach that doesn't introduce too much indirection. :)

svozza: Cool. I was going to put together a PR to make the tests for the other types look like yours so I may as well roll it into that.

davidchambers: I recommend keeping the pull requests separate, in case the convenience function is contentious. ;)

svozza: Haha. Fair enough.

PR #25, opened 2016-01-21
types: define RegexFlags

davidchambers:
This type will be used in the definition of [`S.regex`][1]. It's unlikely to be useful elsewhere, but there's value in defining "[well-defined][2]" types in one place.

[1]: https://github.com/plaid/sanctuary/pull/107
[2]: https://github.com/ramda/ramda/pull/1264#issuecomment-120645134

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -977,6 +977,23 @@ describe('def', function() {
     eq(dec(new Number(-1)), -2);
   });
 
+  it('supports the "RegexFlags" type', function() {
+    eq($.RegexFlags.test(''), true);
+    eq($.RegexFlags.test('g'), true);
+    eq($.RegexFlags.test('i'), true);
+    eq($.RegexFlags.test('m'), true);
+    eq($.RegexFlags.test('gi'), true);
+    eq($.RegexFlags.test('gm'), true);
+    eq($.RegexFlags.test('im'), true);
+    eq($.RegexFlags.test('gim'), true);
+    //  String objects are not acceptable.
+    eq($.RegexFlags.test(new String('')), false);
+    //  Flags must be alphabetically ordered.
+    eq($.RegexFlags.test('mg'), false);
+    //  "Sticky" flag is not acceptable.
+    eq($.RegexFlags.test('y'), false);
+  });
+
   it('uses R.toString-like string representations', function() {
     //  f :: Null -> Null
     var f = def('f', {}, [$.Null, $.Null], function(x) { return x; });
--- a/README.md
+++ b/README.md
@@ -316,6 +316,23 @@ $.RegExp :: Type
 
 Type comprising every RegExp value.
 
+#### `RegexFlags`
+
+```haskell
+$.RegexFlags :: Type
+```
+
+Type comprising the canonical RegExp flags:
+
+  - `''`
+  - `'g'`
+  - `'i'`
+  - `'m'`
+  - `'gi'`
+  - `'gm'`
+  - `'im'`
+  - `'gim'`
+
 #### `String`
 
 ```haskell
--- a/index.js
+++ b/index.js
@@ -284,7 +284,7 @@
   };
 
   //  EnumType :: [Any] -> Type
-  $.EnumType = function(members) {
+  var EnumType = $.EnumType = function(members) {
     var reprs = map(members, show);
     return {
       type: 'ENUM',
@@ -454,6 +454,9 @@
     'sanctuary-def/NonZeroInteger',
     function(x) { return Integer.test(x) && x != 0; }
   );
+
+  //  RegexFlags :: Type
+  $.RegexFlags = EnumType(['', 'g', 'i', 'm', 'gi', 'gm', 'im', 'gim']);
 
   //  arity :: (Number, Function) -> Function
   var arity = function(n, f) {


wmrowan:
LGTM

test/index.js
```diff
@@ -977,6 +977,24 @@ describe('def', function() {
...
+    eq($.RegexFlags.test('i'), true);
+    eq($.RegexFlags.test('m'), true);
+    eq($.RegexFlags.test('gi'), true);
+    eq($.RegexFlags.test('gm'), true);
+    eq($.RegexFlags.test('im'), true);
+    eq($.RegexFlags.test('gim'), true);
+    //  String objects are not acceptable.
+    eq($.RegexFlags.test(new String('')), false);
+    //  Flags must be alphabetically ordered.
+    eq($.RegexFlags.test('mg'), false);
```
kedashoe: Is this preferable or just easier to define this way?

davidchambers: Preferable, in my view. I see `RegexFlags` as a union type:

```haskell
data RegexFlags = Empty | G | I | M | GI | GM | IM | GIM
```

Strings happen to be the best way to achieve this in JavaScript, but I'd like us to think of `'gm'` as we think of `GM` in the Haskell definition above. `MG` is not a constructor in the Haskell definition; similarly `'mg'` is not a member of `RegexFlags`.

Do you find this argument compelling?

kedashoe: I think it would be surprising to users that flags required a certain order. I would find it a bit strange if `grep -rin` got mad at me, or `'xi XI'.replace(/xi/ig, 'hello')` threw

davidchambers: Hmm, that's a fair point.

I do tolerate—and often appreciate—looseness in the tools I use, whereas I appreciate strictness in languages and libraries. I appreciate argument juggling when using `ack`, for example, but rail against it in JavaScript libraries.

Another option is to export `G = 1, I = 2, M = 4`, and define `RegexFlags` as `{0, 1, 2, 3, 4, 5, 6, 7}`. This would avoid the order problem as one could write `G | M` or `M | G`. But where would `G`, `I`, and `M` be defined? Perhaps `S.GLOBAL`, `S.IGNORECASE`, and `S.MULTILINE`, leading to `S.GLOBAL | S.MULTILINE`. This is perhaps slightly clearer—but much less convenient—than `'gm'`.

Hypothetical question: Were we to define `S.G`, `S.I`<sup>†</sup>, `S.M`, `S.GI`, etc., would you expect us to define both `S.GM` and `S.MG`?

I'd love to hear from others. Kevin's concern is legitimate, and I'm not convinced that my current position is optimal.

---

† Ignore the fact that `S.I` already exists. This is just a though exercise. :)

joneshf: I don't think I have a strong opinion one way or the other. Both arguments sounds reasonable.

svozza: You could also define it in terms of an `EnumType` whose the members are all possible combinations.

davidchambers: Great point, Stefano!

```diff
diff --git a/index.js b/index.js
index 6c1dde7..2113362 100644
--- a/index.js
+++ b/index.js
@@ -284,7 +284,7 @@
   };
 
   //  EnumType :: [Any] -> Type
-  $.EnumType = function(members) {
+  var EnumType = $.EnumType = function(members) {
     var reprs = map(members, show);
     return {
       type: 'ENUM',
@@ -456,10 +456,7 @@
   );
 
   //  RegexFlags :: Type
-  $.RegexFlags = NullaryType(
-    'sanctuary-def/RegexFlags',
-    function(x) { return typeof x === 'string' && /^g?i?m?$/.test(x); }
-  );
+  $.RegexFlags = EnumType(['', 'g', 'i', 'm', 'gi', 'gm', 'im', 'gim']);
 
   //  arity :: (Number, Function) -> Function
   var arity = function(n, f) {
diff --git a/test/index.js b/test/index.js
index b4c8de3..79923a1 100644
--- a/test/index.js
+++ b/test/index.js
@@ -978,7 +978,6 @@ describe('def', function() {
   });
 
   it('supports the "RegexFlags" type', function() {
-    eq($.RegexFlags.name, 'sanctuary-def/RegexFlags');
     eq($.RegexFlags.test(''), true);
     eq($.RegexFlags.test('g'), true);
     eq($.RegexFlags.test('i'), true);
```

:zap:

svozza: This would also allow you to support `ig`/`mg` et all should you so wish. I'm not sure if I see the benefit tbh.

davidchambers: @rbuckheit and I had a good discussion about this matter yesterday. I didn't convince Ryan of my position, but I did increase my own conviction. I plan to merge this pull request after my yoga class, but before hitting the green button I'll explain my reasoning as clearly as I can.

davidchambers:
I've decided to merge this pull request, despite the legitimate concern raised by @kedashoe.

Earlier, I wrote:

> I see `RegexFlags` as a union type:
>
> ```haskell
> data RegexFlags = Empty | G | I | M | GI | GM | IM | GIM
> ```

Were we to define `S.regexFlags.gm`, I do not believe anyone would expect `S.regexFlags.mg` as an alias. There's no compelling reason to provide two identical but differently named constructors. If one squints and sees `'gm'` as a nullary data constructor for the RegexFlags type, one would be surprised if `'mg'` were *also* a constructor.

> I think it would be surprising to users that flags required a certain order.

First, I argue that the notion of order applies to `'gm' :: String` but not to `'gm' :: RegexFlags`. In the latter case `'gm'` is a name rather than a sequence of characters.

Secondly, and more importantly, the concern is largely addressed by Stefano's suggestion that we define `RegexFlags` via `EnumType`. An invalid value such as `'mg'` will result in a descriptive exception:

```javascript
//    regex :: RegexFlags -> String -> RegExp
const regex =
def('regex',
    {},
    [$.RegexFlags, $.String, $.RegExp],
    (flags, source) => new RegExp(source, flags));

regex('gm', '^');
// => /^/gm

regex('mg', '^');
// ! TypeError: ‘regex’ expected a value of type ("" | "g" | "i" | "m" | "gi" | "gm" | "im" | "gim") as its first argument; received "mg"
```

kedashoe:
I'm a fan of the principle of least astonishment, but if you're going to surprise me, a good error message is a great mollifier :) And while I still can't say I'm in favor of this, I was never that against it anyway :+1:

davidchambers:
The other thing is we always have the option of expanding the type in a future version. ;)

PR #26, opened 2016-01-23
Add convenience function for boxed types

svozza:
As mentioned in #24. I've only refactored the tests for `ValidNumber` given the controversial nature of this PR. :wink:

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -23,6 +23,11 @@ var errorEq = R.curry(function(type, message, error) {
   return error.constructor === type && error.message === message;
 });
 
+var testBoxed = R.curry(function(type, val, result) {
+  eq(type.test(val), result);
+  eq(type.test(Object(val)), result);
+});
+
 
 var def = $.create($.env);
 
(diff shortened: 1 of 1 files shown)


test/index.js
```diff
@@ -23,6 +23,26 @@ var errorEq = R.curry(function(type, message, error) {
...
+      break;
+    case 'number':
+      boxed = new Number(val);
+      break;
+    default:
+      throw Error('‘testBoxed‘ only accepts primitives; got ' +
+                  typeof val + '.');
+  }
+  eq(type.test(val), result);
+  eq(type.test(boxed), result);
```
davidchambers: Would this remove the need for the `switch` statement?

```diff
-  eq(type.test(boxed), result);
+  eq(type.test(Object(val)), result);
```

svozza: Oh nice, that's a neat little trick.

davidchambers:
This pull request makes two changes:

  - it replaces the verbose—but perhaps more rigorous—`assert.throws`-style assertions with `eq(Type.test(x), false)`-style assertions; and

  - it introduces `testBoxed` to reduce repetition when testing both primitives and their boxed counterparts for membership in a type.

I'd like to see these changes made in separate pull requests. I want to ensure we don't overvalue the second change as a result of seeing it intermingled with the first change. Stefano, would you mind opening a separate pull request for the first change?

svozza:
Only problem with that is if I remove the tests that use `testBoxed`, I get a linting error about defining an unused function.

davidchambers:
How do you feel about this pull request, Stefano? If you're excited about it, please update it. If not, let's close it.

svozza:
I can't say I'm excited about it so I think we can close.

PR #28, opened 2016-01-28
make type checking optional

### LINKED ISSUE #6: option to disable type checking
Several people have suggested making it possible to disable type checking (to reduce the performance impact of sanctuary-def in production, perhaps). I agree that this would be useful.

One possibility is to disable type checking when `process.env.NODE_ENV === 'production'`, though this is not necessarily applicable in the browser, and switching behaviour based on the value of a "shared" environment variable can be problematic. Another option is to disable type checking when `process.env.SANCTUARY_TYPE_CHECK === 'false'`. Neither of these options appeals to me.

We could instead let the user decide whether or not type checking should be performed. The user could then use `process.env.NODE_ENV === 'production'` or any other check. Here's how it might work in practice:

```javascript
// Always perform type checking.
const def = $.create(true, $.env);
```

```javascript
// Never perform type checking.
const def = $.create(false, $.env);
```

```javascript
// Only perform type checking in "development" mode.
const def = $.create(process.env.NODE_ENV === 'development', $.env);
```

What do you think of this proposal? Should the first argument to `$.create` have type `{ typeCheck :: Boolean }` rather than `Boolean`? If so, should `$.create` take a single argument of type `{ typeCheck :: Boolean, types :: [Type] }`? Is there a completely different way to make this behaviour configurable?

davidchambers:
Closes #6

I'm not fond of Boolean flags, but `{ checkTypes :: Boolean }` would introduce a separate problem: one could provide an invalid value such as `{typeCheck: true}`. I'd prefer to avoid having to name the option.

When viewing the changes, run `location.search = '?w='` in your console for best results. ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,15 @@ const env = $.env.concat([Integer, NonZeroInteger]);
 The next step is to define a `def` function for the environment:
 
 ```javascript
-const def = $.create(env);
+const def = $.create(true, env);
+```
+
+The first argument to `$.create` determines whether type checking is enabled.
+This allows one to only pay the performance cost of run-time type checking
+during development. For example:
+
+```javascript
+const def = $.create(process.env.NODE_ENV === 'development', env);
 ```
 
 `def` is a function for defining functions. For example:
@@ -650,7 +658,10 @@ For example:
 //    TimeUnit :: Type
 const TimeUnit = $.EnumType(['milliseconds', 'seconds', 'minutes', 'hours']);
 
-const def = $.create($.env.concat([TimeUnit, $.ValidDate, $.ValidNumber]));
+//    env :: [Type]
+const env = $.env.concat([TimeUnit, $.ValidDate, $.ValidNumber]);
+
+const def = $.create(true, env);
 
 //    convertTo :: TimeUnit -> ValidDate -> ValidNumber
 const convertTo =
--- a/index.js
+++ b/index.js
@@ -671,8 +671,8 @@
     );
   };
 
-  //  create :: [Type] -> Function
-  $.create = function(_env) {
+  //  create :: (Boolean, [Type]) -> Function
+  $.create = function(checkTypes, _env) {
     //  env :: [Type]
     var env = map(_env, function(x) {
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -30,7 +30,15 @@ const env = $.env.concat([Integer, NonZeroInteger]);
 The next step is to define a `def` function for the environment:
 
 ```javascript
-const def = $.create(env);
+const def = $.create(true, env);
+```
+
+The first argument to `$.create` determines whether type checking is enabled.
+One may elect to pay the performance cost of run-time type checking only during
+development. For example:
```
davidchambers: It's possible to misread the second sentence:

_**Only during development** may one elect to pay the performance cost of run-time type checking._

:100: points to the first person to suggest an unambiguous alternative. :)

svozza: How about:

*This allows one to only pay the performance cost of run-time type type checking during development.*

davidchambers: :100: points for Stefano!

svozza: You're up late!

README.md
```diff
@@ -30,7 +30,15 @@ const env = $.env.concat([Integer, NonZeroInteger]);
...
 
 ```javascript
-const def = $.create(env);
+const def = $.create(true, env);
+```
+
+The first argument to `$.create` determines whether type checking is enabled.
+This allows one to only pay the performance cost of run-time type checking
+during development. For example:
+
```
soroushjp: Great to make this an argument rather than env var or an arbitrary check.

soroushjp:
@davidchambers Should we add some sort of CI to automatically run tests on PRs?

README.md
davidchambers: Yep. I want to stay out of the application configuration game. :)

davidchambers:
> Should we add some sort of CI to automatically run tests on PRs?

We're using [CircleCI][1].


[1]: https://circleci.com/gh/plaid/sanctuary

soroushjp:
> We're using CircleCI.

Sorry, missed that -- thought the "1 successful check" was referring to merge conflicts.

Issue #7, opened 2015-12-23
Possible to give Record types a name property?


Wideshanks:
So instead of printing a list of all the fields as in:

https://github.com/plaid/sanctuary-def/blob/master/index.js#L293

--instead print the name of the Record in TypeError messages?

davidchambers:
This is certainly worth discussing. I'd like to understand your motivations, as I have some questions:

  - Are you concerned about the verbosity of error messages for complex record types?

  - Should it even be valid to define two equivalent record types in an environment? For example:

    ```javascript
    const Coord = $.RecordType({x: $.Number, y: $.Number});
    const Point = $.RecordType({x: $.Number, y: $.Number});

    const def = $.create($.env.concat([Coord, Point]));
    ```

My thinking at the time was that `{ age :: Number, name :: String }` is more descriptive than `Person`. If an error message just referenced `Person`, one might need to consult code or documentation in order to learn the shape of a Person.

Wideshanks:
> concerned about the verbosity of error messages for complex record types

Yeah, verbosity in (very) large records:

![scrolling-output](https://cloud.githubusercontent.com/assets/2061925/12061094/94c13966-af4b-11e5-9ac9-df1287b2cf8b.png)

Or maybe I should avoid using records when the data structures are that big?

davidchambers:
> Or maybe I should avoid using records when the data structures are that big?

This seems to me like perfectly reasonable use of a record type. :)

It's interesting to note that in this case the representation of the record type accounts for only about 5% of the error message: the vast majority is devoted to the representation of the invalid return value.

I wonder whether the real issue is that the error messages do not explain *why* a value is invalid. Consider this example:

```javascript
const Point = $.RecordType({x: $.Number, y: $.Number});

const def = $.create($.env.concat([Point]));

//    dist :: Point -> Point -> Number
const dist =
def('dist',
    {},
    [Point, Point, $.Number],
    (a, b) => Math.sqrt(Math.pow(a.x - b.x, 2) + Math.pow(a.y + b.y, 2)));

dist({x: 0, y: 0}, {x: 3});
// ! TypeError: ‘dist’ expected a value of type { x :: Number, y :: Number } ↩︎
//              as its second argument; received {"x": 3}

dist({x: 0, y: 0}, {x: 3, y: '4'});
// ! TypeError: ‘dist’ expected a value of type { x :: Number, y :: Number } ↩︎
//              as its second argument; received {"x": 3, "y": "4"}
```

[...]

PR #23, opened 2016-01-18
multi-line error messages

kedashoe:
Before
```
TypeError: ‘aa’ expected a value of type Number as its second argument; received /x/
```
After
```
TypeError: ‘aa’ expected its first and second arguments to be of the same type; the first is of type Number, however the second, /x/, is of type RegExp
```
I realize one might normally file an issue first, but was just playing around with the code and figured may as well submit a PR, no hard feelings if you don't want to go down this path. Couple notes

- When the placeholder is used, the message is a little funny (see https://github.com/kedashoe/sanctuary-def/blob/same-type-var-error/test/index.js#L974-L976). I was going to address it but left it as it was as it actually conveys a little more information
- This hit 8 tests, so I didn't add any new ones
- The pattern `commonTypes(map(values, determineActualTypes))` is now used 4 times, and I believe this is the only time `determineActualTypes` is used, but I just reused the pattern rather than refactor
- I liked `however` in the error message at first, but don't anymore. I feel like maybe there is a really nice multi-line error message to be arrived at somewhere here. (And not to hijack my own PR, but what do you all think about using `1st`, `2nd`... rather than `first`, `second`...? to me that is much easier to read and quickly process)

Great work on the lib :)

PR #30, opened 2016-02-10
list reasons when a value is not a member of the specified record type

davidchambers:
As discussed in #7

Example:

```javascript
//    Point :: Type

const def = $.create(true, $.env.concat([Point]));

//    dist :: Point -> Point -> Number
const dist =
def('dist',
    {},
    [Point, Point, $.Number],
    (p, q) => Math.sqrt(Math.pow(p.x - q.x, 2) + Math.pow(p.y - q.y, 2)));

dist({x: 1, y: 1}, {x: 4, y: 5});
// => 5

dist({x: '1'});
// ! TypeError: Invalid value
//
//   dist :: { x :: Number, y :: Number } -> { x :: Number, y :: Number } -> Number
//           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
//                        1
//
//   1)  {"x": "1"} :: Object
//
//   The value at position 1 is not a member of { x :: Number, y :: Number }.
//
//   Type checking failed for the following reasons:
//
//     - Value of ‘x’ field is not a member of expected type, Number
//
//     - Missing ‘y’ field
```

Even if the first line is unhelpful due to verbose string representations, the additional information will highlight the problems with the value.

Are you happy with the structure and phrasing of these messages?

/cc @kedashoe, @Wideshanks

### DIFF
--- a/index.js
+++ b/index.js
@@ -793,17 +793,27 @@
       );
     }
 
+    //  validate :: Any -> [String]
+    var validate = function(x) {
+      if (typeOf(x) !== 'Object') return [show(x) + ' is not a record'];
+      var errors = [];
+      for (var idx = 0; idx < names.length; idx += 1) {
+        var name = names[idx];
+        if (!has(name, x)) {
+          errors.push('Missing ‘' + name + '’ field');
+        } else if (!fields[name].test(x[name])) {
+          errors.push('Value of ‘' + name + '’ field is not a member of ' +
+                      'expected type, ' + show(fields[name]));
+        }
+      }
+      return errors;
+    };
+
     return {
       '@@type': 'sanctuary-def/Type',
       type: 'RECORD',
-      test: function(x) {
-        if (x == null) return false;
-        for (var idx = 0; idx < names.length; idx += 1) {
-          var name = names[idx];
-          if (!has(name, x) || !fields[name].test(x[name])) return false;
-        }
-        return true;
-      },
+      test: function(x) { return isEmpty(validate(x)); },
+      validate: validate,
       toString: function() {
         var s = '{';
         for (var idx = 0; idx < names.length; idx += 1) {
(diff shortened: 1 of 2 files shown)


kedashoe:
I love it, but I think this will take a bit of work to merge with #23? `invalidArgument` is only called now in `def` and `invalidReturnValue` is removed altogether, so this would have to be worked into `invalidValue`. Are you thinking we could give types the opportunity to explain why they are an invalid (eg your `validate`)? We can just change https://github.com/plaid/sanctuary-def/pull/23/files#diff-168726dbe96b3ce427e7fedce31bb0bcR1150 to check for it. So maybe not that much work afterall? :) End up with something like

```
Invalid value

//    dist :: Point -> Point -> Number
              ^^^^^
                1

1)  Object :: {"x": "1"}

The value at position 1 is not a member of { x :: Number, y :: Number }. Type checking failed for the following reasons:

  - Value of ‘x’ field is not a member of expected type, Number
  
  - Missing ‘y’ field
```

davidchambers:
> I love it, but I think this will take a bit of work to merge with #23?

I won't merge this until we've merged #23. :)

> Are you thinking we could give types the opportunity to explain why they are an invalid (eg your `validate`)?

This would potentially be very useful. From a usability perspective I appreciate the straightforwardness of `Any -> Boolean`, but the idea of using `Any -> [String]` for all types—rather than having what amounts to a special case for record types—is appealing.

One thing that became apparent to me while using this branch today is that the extra information is of limited value if the approach is not recursive. We're hoping to solve the information overload that results from complex, nested record types, but the current approach just pushes the problem down a level.

davidchambers:
I've rebased this pull request, but I'd still like to make the code recursive before merging.

kedashoe:
Any thought to providing a name for record types? eg `var Point = $.RecordType('Point', {x: $.Number, y: $.Number});`. I could then see something like
```
Invalid value

length :: Line -> Number
          ^^^^
           1

1)  Line{ "start": Point{ "x": "a", "y": "b" } } :: Object

The value at position 1 is not a member of Line.

Type checking failed for the following reasons:

  - Value of ‘start’ field is not a member of expected type, Point{ x :: Number, y :: Number }
```
where all valid fields are omitted (or maybe elided? although I think that may just turn into noise).

davidchambers:
> Any thought to providing a name for record types?

This is quite appealing now that we display type signatures in error messages. It's certainly something for us to consider.

davidchambers:
Superseded by #55

PR #31, opened 2016-02-10
validate argument to RecordType constructor

davidchambers:
@soroushjp wrote some code earlier today which exploded and was difficult to diagnose. He mistakenly used `{…}` rather than `$.RecordType({…})` as a value in an object provided to `$.RecordType`. This pull request adds a check for such errors, making them easy to diagnose and correct.

### DIFF
--- a/index.js
+++ b/index.js
@@ -198,6 +198,7 @@
 
   //  Any :: Type
   var Any = {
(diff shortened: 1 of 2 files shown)


soroushjp:
LGTM @davidchambers

PR #33, opened 2016-02-17
define StrMap type constructor

### LINKED ISSUE #29: StrMap type
It would be a useful to have a StrMap type for values such as `{x: 'X', y: 'Y', z: 'Z'}` and `{x: 1, y: 2, y: 3}`. It would be a unary type, like [`$.Array`][1]. `$.StrMap(a)` would satisfy the requirements of [Functor][2], unlike `$.Object` which does not guarantee a common value type.

[1]: https://github.com/plaid/sanctuary-def#array
[2]: https://github.com/fantasyland/fantasy-land#functor

davidchambers:
Closes #29

```haskell
StrMap :: Type -> Type
```

I thought this would be straightforward. _It's just another unary type, like [`$.Array`][1]!_ No such luck.

The key difference between `[1, 2, 'XXX']` and `{a: 1, b: 2, c: 'XXX'}` is that the former is not a member of any type in `$.env`, whereas the latter is a member of [`$.Object`][2]. All `$.Array.test` must do is determine whether the value is an array, since there's machinery elsewhere for determining the types of a container's values. Were we to use the same approach for `$.StrMap`, `{a: 1, b: 2, c: 'XXX'}` would be considered an invalid value, because the machinery would fail to find a type of which `1`, `2`, and `'XXX'` are all members.

In order to determine whether a value is a member of `StrMap a`, we must first determine whether it's an object. If it is, we must then verify that there's at least one type in the environment of which all its values are members. Since we require access to the environment, `$.StrMap.test` cannot answer our question (without an undesirable API change). As a result we cannot simply rely on dynamic dispatch: instead, we must use the newly defined `test` function which contains a special case for `'sanctuary-def/StrMap'`.

I did warn that it wouldn't be pretty. ;)

[1]: https://github.com/plaid/sanctuary-def#array
[2]: https://github.com/plaid/sanctuary-def#object

### DIFF
--- a/index.js
+++ b/index.js
@@ -432,6 +432,15 @@
     function(nullable) { return nullable === null ? [] : [nullable]; }
   );
 
+  //  StrMap :: Type -> Type
+  var StrMap = UnaryType(
+    'sanctuary-def/StrMap',
+    function(x) { return $.Object.test(x); },
+    function(strMap) {
+      return map(keys(strMap), function(k) { return strMap[k]; });
+    }
+  );
+
   //  $$type :: a -> String
   var $$type = function(x) {
     return x != null && typeOf(x['@@type']) === 'String' ?
@@ -468,6 +477,7 @@
     ($.Number     = type0('Number')),
     ($.Object     = type0('Object')),
     ($.RegExp     = type0('RegExp')),
+    ($.StrMap     = StrMap),
     ($.String     = type0('String')),
     ($.Undefined  = type0('Undefined'))
   ];
@@ -772,7 +782,7 @@
       'Definition of ' + LEFT_SINGLE_QUOTATION_MARK + name + RIGHT_SINGLE_QUOTATION_MARK +
         ' references ' + type.name + ' which is not in the environment:',
       '',
-      map(chain(env, rejectAny), prefix('  - ')).join('\n')
+      map(map(chain(env, rejectAny), showType), prefix('  - ')).join('\n')
     ])));
   };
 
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -701,6 +723,73 @@
...
+  };
+
+  //  consistentTypes :: [Type] -> [Any] -> [Type]
+  var consistentTypes = function(env) {
+    return function(values) {
+      return chain(commonTypes(map(values, determineActualTypes(env))),
+                   rejectInconsistent);
+    };
+  };
+
```
davidchambers: This function is similar to `valueTypes` defined in #23.

index.js
```diff
@@ -701,6 +723,74 @@
...
+            return true;
+          case 'NULLARY':
+            return t.test(x);
+          case 'UNARY':
+            var xs = t._1(x);
+            return t.name === 'sanctuary-def/StrMap' ?
+              ($.Object.test(x) &&
+               (isEmpty(xs) ||
+                !isEmpty(consistentTypes(env)(xs)) && all(xs, recur(t.$1)))) :
+              t.test(x) && all(xs, recur(t.$1));
```
davidchambers: This is pretty awful, but I don't think it can be simplified. To be a member of `sanctuary-def/StrMap`, a value must:

  - satisfy `$.Object.test`; and
  - either:
      - be empty; or
      - have values which are all members of at least one type and which a members of `t.$1`.

Actually, it seems the last point could be separated into two: one for type variables; one for other types.

svozza: Should `all(xs, recur(t.$1))` not be on a new line?

index.js
```diff
@@ -701,6 +723,74 @@
...
+          case 'VARIABLE':
+            return true;
+          case 'NULLARY':
+            return t.test(x);
+          case 'UNARY':
+            var xs = t._1(x);
+            return t.name === 'sanctuary-def/StrMap' ?
+              ($.Object.test(x) &&
+               (isEmpty(xs) ||
+                !isEmpty(consistentTypes(env)(xs)) && all(xs, recur(t.$1)))) :
```
svozza: Should `all(xs, recur(t.$1))` not be on a new line?

index.js
```diff
@@ -362,6 +350,20 @@
     function(nullable) { return nullable === null ? [] : [nullable]; }
   );
 
+  //  StrMap :: Type -> Type
+  var StrMap = UnaryType(
+    'sanctuary-def/StrMap',
+    function(x) {
```
svozza: What's going on here?

test/index.js
```diff
@@ -862,6 +865,97 @@ describe('def', function() {
...
+                   'first argument; received {"a": 1, "b": "XXX"}'));
+
+    //  testUnaryType :: [StrMap Number] -> [StrMap Number]
+    var testUnaryType =
+    def('testUnaryType',
+        {},
+        [$.Array($.StrMap($.Number)), $.Array($.StrMap($.Number))],
+        R.identity);
+
+    eq(testUnaryType([{x: 1}, {y: 2}, {z: 3}]), [{x: 1}, {y: 2}, {z: 3}]);
```
svozza: Do we want to test `StrMaps` created with `Object.create(null)`?

davidchambers: Good idea. I'll add some assertions.

svozza:
I find the code easier to follow with the new changes. :+1:

test/index.js
davidchambers: :zap:

svozza:
LGTM.

index.js
```diff
@@ -1051,40 +1061,46 @@
...
-        if (!isEmpty(common)) return common;
-        //  If none of the values is a member of a type in the environment,
-        //  and all the values have the same type identifier, the values are
-        //  members of a "foreign" type.
-        if (isEmpty(filterTypesByValues(consistentTypes, values)) &&
-            all(values.slice(1), $$typeEq($$type(values[0])))) {
-          //  Create a nullary type for the foreign type.
-          return [type0($$type(values[0]))];
+    //  _determineActualTypes :: (Boolean, [Object], [Any]) -> [Type]
+    var _determineActualTypes = function recur(loose, $seen, values) {
```
Avaq: Isn't `recur` too general of a function name now that it's sharing scope with others?

index.js
```diff
@@ -1051,40 +1061,46 @@
...
+      if (isEmpty(values)) return [Unknown];
+      //  consistentTypes :: [Type]
+      var consistentTypes = chain(env, rejectAny);
+      //  typeses :: [[Type]]
+      var typeses = map(values, function(value) {
+        if (typeof value === 'object' && value != null) {
```
Avaq: Functions can have references too, and fail this check. Also Regex was considered a function in Chrome 1 to 12.

index.js
davidchambers: This is a [named function expression][1].

> If you want to refer to the current function inside the function body, you need to create a named function expression. This name is then local only to the function body (scope). This also avoids using the non-standard [`arguments.callee`][2] property.


[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/function#Named_function_expression
[2]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments/callee

index.js
davidchambers: Great catch! I've added `|| typeof value === 'function'` to handle these cases.

index.js
Avaq: Oh. I didn't know about that. :+1:

PR #37, opened 2016-03-06
simplify logic for generating error messages

davidchambers:
I'm tremendously grateful to @kedashoe for the descriptive error messages introduced in #23. This library is significantly more useful as a result of that pull request.

But one thing about #23 didn't sit well with me: the extent to which the code relied on mutable state. This pull request removes this mutable state, and sheds several hundred lines of code in the process.

This pull request also changes the formatting of error messages in three ways:

  - The representations of parameterized types (as returned by `showType`) no longer feature parentheses at the top level. `[1, 2, 3] :: (Array Number)`, for example, is now `[1, 2, 3] :: Array Number`. Nested parameterized types are still parenthesized. We are already doing this for parameterized types in type signatures, so it makes sense to apply the same approach to type representations elsewhere.

  - Type representations are now quoted when used in a sentence. For example:

    ```
    The value at position 1 is not a member of ‘Number’.
    ```

    This is valuable in conjunction with the removal of top-level parentheses for parameterized types, as the quotation marks perform the grouping role. `(Either a b)`, for example, is now `‘Either a b’`.

  - The message for type-variable constraint violations now presents `value :: types` pairs as a flat list – not because this is an improvement, but because I discovered a problem with our implementation.

    ```
    Type-variable constraint violation

    defaultTo :: a -> Nullable a -> a
                 ^             ^
                 1             2

    1)  0 :: Number

    2)  String( "XXX" :: String )

    Since there is no type of which all the above values are members, the type-variable constraint has been violated.
    ```

    Note that `String( "XXX" :: String )` should be `Nullable( "XXX" :: String )`.

    I'm certainly open to reinstating the more sophisticated presentation (and fixing the Nullable issue). This pull request is focused on internal simplification. I'm willing to knowingly introduce a regression in exchange for a simpler foundation upon which to add features (including the one removed in this pull request).

### DIFF
--- a/index.js
+++ b/index.js
@@ -69,13 +69,6 @@
     };
   };
 
-  //  functionName :: Function -> String
-  var functionName = function(f) {
-    // String(x => x) evaluates to "x => x", so the pattern may not match.
-    var match = String(f).match(/^function (\w+)/);
-    return match == null ? '' : match[1];
-  };
-
   //  has :: (String, Object) -> Boolean
   var has = function(key, obj) { return hasOwnProperty.call(obj, key); };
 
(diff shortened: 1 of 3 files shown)


kedashoe:
> The representations of parameterized types (as returned by showType) no longer feature parentheses at the top level.

:+1: 

> Type representations are now quoted when used in a sentence.

:+1: 

> Note that String( "XXX" :: String ) should be Nullable( "XXX" :: String ).

This was nothing to do with #23 per se, but with the implementation of `Nullable`. There was no special case for parameterized types, only for handling builtins (just `Array` so far, but once `StrMap` lands I would think we would handle that as well). I don't think we want users having to wrap all their `Nullable` natives, so a special case for `Nullable` could be made. I do think it's a shame to lose this information. For type-variable constraints the user will probably be able to infer the containing types, but for `Invalid Value` errors they will be totally lost.

> I'm willing to knowingly introduce a regression in exchange for a simpler foundation upon which to add features (including the one removed in this pull request).

There was no mutation here, looks like it was about [60 SLOC](https://github.com/plaid/sanctuary-def/blob/master/index.js#L1273-L1339) to build and [30 SLOC](https://github.com/plaid/sanctuary-def/blob/master/index.js#L1131-L1171) to show the error values. Unless you find that work unsalvageable, I don't know that we've a simpler foundation, but just a much simpler problem we've solved (I am certain though that code *can* be simplified, better function names, some functions broken down more, etc). I feel like it might be better to handle `Nullable` and clean things up a bit rather than to lose everything (I was pretty happy with this part, but maybe I am wrong) and kick that can down the road.

> sheds several hundred lines of code in the process.

I do think the type signature stuff ended up over-engineered, nice work!

index.js
```diff
@@ -1128,90 +703,183 @@
...
+      _(name + ' :: ') + carets + _(' => ') +
+      padding + r('^')(typeVarName) + '\n' +
+      _(name + ' :: ' + carets + ' => ') +
+      padding + label('1')(typeVarName) + '\n\n' +
+      '1)  ' + showValueAndType([value, actualTypes]) + '\n\n' +
+      LEFT_SINGLE_QUOTATION_MARK + name + RIGHT_SINGLE_QUOTATION_MARK +
+      ' requires ' +
+      LEFT_SINGLE_QUOTATION_MARK + typeVarName + RIGHT_SINGLE_QUOTATION_MARK +
+      ' to satisfy the ' + typeClass + ' type-class constraint;' +
+      ' the value at position 1 does not.\n'
```
kedashoe: What do you think about adding

`// jscs:disable maximumLineLength`

to the multi-line error message functions so we could write something like

```js
return new TypeError(
  'Type-class constraint violation\n' +
  '\n' +
  name + ' :: ' + constraintsRepr(constraints) + showTypeSig(expTypes) + '\n' +
  _(name + ' :: ') + carets + _(' => ') + padding + r('^')(typeVarName) + '\n' +
  _(name + ' :: ' + carets + ' => ') + padding + label('1')(typeVarName) + '\n' +
  '\n' +
...
```
For me this is a lot easier to read/understand, similar to how we lined up the error messages in `index/test.js`

davidchambers: Great idea! This significantly improves readability. :zap:

index.js
```diff
@@ -1270,78 +947,15 @@
...
-        isParameterizedType: false,
-        types: valueTypes(values),
-        values: values
-      };
-    };
-
-    //  typesAndValues :: (type, [Any]) -> ErrorRepr
-    var typesAndValues = function(t, values) {
-      if (isParameterizedType(t)) {
-        var constructorName = functionName(values[0].constructor);
```
davidchambers: This has a few shortcomings:

```javascript
//    Identity :: a -> Identity a
const Identity = x => ({'@@type': 'my-package/Identity', value: x});

Identity(42).constructor.name;
// => 'Object'

S.Just(42).constructor.name;
// => 'Maybe'
```

We'd like to use `Identity( … )` and `Just( … )` in these cases to provide context, but I don't see how we can do so.

kedashoe: > Identity(42).constructor.name;
> // => 'Object'

Could check for `@@type` if constructor is `Object`?

> S.Just(42).constructor.name;
> // => 'Maybe'

The way that Maybe -> Just/Nothing and Either -> Left/Right are defined is pretty unique I would say. It seems pretty odd to have a pseudo base class that is always checking `isJust`, `
Another benefit of this approach is that the error messages will refer to Either#foo rather than Left#foo/Right#foo

kedashoe: > Identity(42).constructor.name;
> // => 'Object'

Could check for @@type if constructor.name is Object?

> S.Just(42).constructor.name;
> // => 'Maybe'

The way that Maybe -> Just/Nothing and Either -> Left/Right are defined is pretty unique I would say. In the comments to plaid/sanctuary#123 you [wrote](https://github.com/plaid/sanctuary/pull/123#issuecomment-168406709):

> When defining a method for both Left and Right, say, it's convenient to share the constraints and the list of types, so it seems more appropriate to have a single definition with one internal branch than two separate definitions with no branches.

Another alternative is to partially apply `def` (and `method`), leaving off only the function definition. So take `Either` and `chain` for example:

```js
var Either_chain = method(
  'Either#chain',
  {},
  [$Either(a, b), $.Function, $Either(a, c)]
);

Right.prototype.chain = Either_chain(function(right, f) {
  return f(right.value);
});

Left.prototype.chain = Either_chain(function(left, f) {
  return left;
});
```
Maybe the best of both worlds? We don't need to `def` functions in multiple places and the actual function bodies become super simple again, without any type branching.

We could then go back to the more standard constructors before #123, the instances will actually be instances of the things they are instantiating, our error messages will tell us what we actually have, etc. And of course we don't need to bring back inheritance, although we could if we wanted?

davidchambers: These are the three issues with which we're faced:

```javascript
Nullable a          'XXX'               String( "XXX" :: String )
         ^

Identity a          Identity('XXX')     Object( "XXX" :: String )
         ^

Either a b          Left('XXX')         Either( "XXX" :: String )
       ^
```

This first could be fixed with a special case for nullable types. I've no problem with this as nullable types are weird and already require special treatment to avoid infinite recursion when determining a value's types.

The second issue could be fixed, as you suggest, by using the unprefixed type identifier (which could come from the type, I believe, so we needn't rely on the `@@type` property). For types with multiple data constructors, this would bring the third issue into play.

The third issue could be fixed, as you suggest, by requiring that all values provide a suitable `constructor.name`. This doesn't sit well with me. It suggests that values should be created via object-oriented constructor functions. This may lead people to view Left and Right, say, as subtypes of Either, an abstract base type. Inheritance muddies the conversation.

The other problem with constructor functions is they facilitate the use of `instanceof`, which I consider harmful. One could, of course, include `constructor: {name: 'Identity'}` in a value to "trick" sanctuary-def, but this strikes me as quite confusing.

I'd prefer to consult another `@@something` property than to use `constructor.name`.

It appears not to be possible to display context *consistently* without introducing another convention (such as `constructor.name` or `@@something`). Given that we already underline the relevant type variables, I don't think a new convention is warranted. What do you think, @kedashoe?

kedashoe: That all makes sense, I agree a new convention is not warranted.

PR #38, opened 2016-03-13
support values of "foreign" types

davidchambers:
See plaid/sanctuary#165

I'd love a review from @soroushjp or anyone else in the community. :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -69,6 +69,17 @@
     };
   };
 
+  //  filter :: ([a], (a -> Boolean)) -> [a]
+  var filter = function(xs, pred) {
+    var result = [];
+    for (var idx = 0; idx < xs.length; idx += 1) {
+      if (pred(xs[idx])) {
+        result.push(xs[idx]);
+      }
+    }
+    return result;
+  };
+
   //  has :: (String, Object) -> Boolean
   var has = function(key, obj) { return hasOwnProperty.call(obj, key); };
 
@@ -249,6 +260,14 @@
     toString: always('???')
   };
 
+  //  Inconsistent :: Type
+  var Inconsistent = {
+    '@@type': 'sanctuary-def/Inconsistent',
+    type: 'INCONSISTENT',
+    test: K(false),
+    toString: always('???')
+  };
+
   //  TypeVariable :: String -> Type
   $.TypeVariable = function(name) {
     return {
@@ -290,7 +309,7 @@
     };
   };
 
-  //  UnaryType.from :: Type -> Type
+  //  UnaryType.from :: Type -> (Type -> Type)
   UnaryType.from = function(t) {
     return UnaryType(t.name, t.test, t._1);
   };
@@ -319,7 +338,7 @@
     };
   };
 
-  //  BinaryType.from :: Type -> Type
+  //  BinaryType.from :: Type -> ((Type, Type) -> Type)
   BinaryType.from = function(t) {
     return BinaryType(t.name, t.test, t._1, t._2);
   };
@@ -598,8 +617,8 @@
     }
   };
 
-  //  rejectInconsistent :: Type -> [Type]
-  var rejectInconsistent = function recur(t) {
+  //  rejectAny :: Type -> [Type]
+  var rejectAny = function recur(t) {
     switch (t.type) {
       case 'ANY':
         return [];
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -999,8 +1004,23 @@
...
               okTypes = valueTypes(values);
-              if (isEmpty(okTypes) && !isEmpty(values)) {
-                throw orphanArgument(env, name, _value, index);
+              if (!isEmpty(values)) {
+                //  If there's at least one value, and none of the values is
+                //  a member of a type in the environment, and all the values
+                //  have the same type identifier, the values are members of
+                //  a "foreign" type.
+                //
+                //  Create a nullary type for the foreign type, for inclusion
```
soroushjp: s/nullary/Nullary

test/index.js
```diff
@@ -1589,6 +1589,33 @@ describe('def', function() {
...
+    //  id :: a -> a
+    var id = def('id', {}, [a, a], R.identity);
+
+    eq(id({'@@type': 'my-package/Foreign'}), {'@@type': 'my-package/Foreign'});
+
+    throws(function() { id([{'@@type': 'my-package/Foo'},
+                            {'@@type': 'my-package/Bar'}]); },
+           errorEq(TypeError,
+                   '‘id’ received [{"@@type": "my-package/Foo"}, ' +
+                   '{"@@type": "my-package/Bar"}] as its first argument, ' +
```
soroushjp: This doesn't seem to be the best possible error message for the user. From their point of view, both `my-package/Foo` and `my-package/Bar` are valid types, but they don't see these types in the list below. I understand that `Array ???` captures the specific `TypeError` here, but it would be quite difficult to discern for most people. Maybe we should have a specific error message for an `Array` with no shared consistent type for all values?

index.js
davidchambers: I intended this to be read as plain English rather than as a reference to `$.NullaryType`.

test/index.js
davidchambers: > Maybe we should have a specific error message for an `Array` with no shared consistent type for all values?

Good idea. It needn't be specific to `$.Array`, either. It could apply to any parameterized type.

In fact, now that I think about it this is really a TVCV (type-variable constraint violation). I'll see whether we can use `constraintViolation` here instead.

soroushjp: Makes sense to me @davidchambers.

davidchambers: This turned out to be much more involved than I anticipated, but I'm very pleased with the results. Have a look at the changes to the test suite, @soroushjp. :)

index.js
```diff
@@ -249,6 +260,14 @@
...
   };
 
+  //  Inconsistent :: Type
+  var Inconsistent = {
+    '@@type': 'sanctuary-def/Inconsistent',
+    type: 'INCONSISTENT',
+    test: K(false),
+    toString: always('???')
+  };
+
```
davidchambers: Whereas `[] :: Array Unknown`, `['foo', false] :: Array Inconsistent`. Using the same internal representation for both types was problematic since we need to be able to differentiate the two cases.

index.js
```diff
@@ -620,21 +639,20 @@
     );
   };
 
-  //  equalTypes :: (Type, Type) -> Boolean
-  var equalTypes = function equalTypes(t1, t2) {
+  //  equalTypes :: (Type, Type, Boolean) -> Boolean
+  var equalTypes = function equalTypes(t1, t2, loose) {
+    if (t1.type === 'INCONSISTENT' || t2.type === 'INCONSISTENT') return loose;
```
davidchambers: We set `loose` to `true` when generating error messages, which allows us to display `['foo', false] :: Array ???` rather than `['foo', false] :: ???`.

index.js
```diff
@@ -645,22 +663,51 @@
...
-      for (var idx = 1; idx < typeses.length; idx += 1) {
-        common = false;
-        for (var idx2 = 0; idx2 < typeses[idx].length; idx2 += 1) {
-          if (equalTypes(t, typeses[idx][idx2])) {
-            common = true;
-            break;
-          }
-        }
+  //  commonTypes :: ([[Type]], Boolean) -> [Type]
+  var commonTypes = function(typeses, loose) {
```
svozza: Can you explain a little what's going on here?

davidchambers: Sure.

Let's say we have `values :: [0, 0.5, 1]`.

This might give `typeses :: [[Number, Integer], [Number], [Number, Integer]]`.

`commonTypes(typeses, false)` would then evaluate to `[Number]`.

Now let's discuss the `loose` parameter.

Let's say we have `values :: ['foo', false]`.

This might give `typeses :: [[String], [Boolean]]`.

`commonTypes(typeses, false)` would then evaluate to `[]`, indicating a type error.

We'd like to display `['foo', false] :: Array ???` rather than `['foo', false] :: ???`. The `loose` parameter allows us to do so. We set `loose` to `true` when determining validity; `false` when generating error messages.

index.js
```diff
@@ -846,9 +882,29 @@
...
+  var conflictingTypeVar = function(name, constraints, expTypes, fst) {
+    var nameAndConstraints = name + ' :: ' + constraintsRepr(constraints);
+    var expTypeRepr = showType(expTypes[fst.index]);
+    var padding = _(_showTypeSig(expTypes.slice(0, fst.index)));
+
+    return new TypeError(unlines([
+      'Type-variable constraint violation',
+      '',
+      nameAndConstraints + showTypeSig(expTypes),
+      _(nameAndConstraints) + padding + trimTrailingSpaces(r('^')(expTypeRepr)),
```
svozza: We should use `left-pad` here. :wink:

index.js
svozza: Thanks David, :)

svozza:
I'm not going to claim I fully grok what's going on here but it this looks fine to me.

davidchambers:
Thank you for reviewing these changes, @svozza. :)

PR #43, opened 2016-04-02
underline nested type variable upon type-class constraint violation

davidchambers:
Before:

    concatMaybes :: Semigroup a => Maybe a -> Maybe a -> Maybe a
                    ^^^^^^^^^^^    ^
                                   1

After:

    concatMaybes :: Semigroup a => Maybe a -> Maybe a -> Maybe a
                    ^^^^^^^^^^^          ^
                                         1

Does this change look okay to you, @kedashoe?

### DIFF
--- a/index.js
+++ b/index.js
@@ -96,6 +96,9 @@
     return result.sort();
   };
 
+  //  last :: [a] -> a
+  var last = function(xs) { return xs[xs.length - 1]; };
+
   //  map :: ([a], (a -> b)) -> [b]
   var map = function(xs, f) {
     var result = [];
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -830,13 +838,14 @@
                       reprs.join(_(', ')));
 
     var padding = _(showTypeSig_(expTypes.slice(0, index)));
```
kedashoe: I believe this line should be something like

`var padding = _(showTypeSig_(expTypes.slice(0, isNaN(index) ? expTypes.length - 1 : index)));`

to properly point to type variables in the return value.

kedashoe: (also might want to add a test for pointing to return values, which is how I discovered something might be amiss..)

index.js
```diff
@@ -830,13 +838,14 @@
...
+    var f = underline(typePath[0])(propPath);
 
     return new TypeError(unlines([
```
kedashoe: I know I already said it, but I love the formatting of these now with the longer lines + `unlines` :)

index.js
davidchambers: Great catch! I'll add a failing test case then make your suggested fix.

index.js
davidchambers: :smile:

index.js
davidchambers: :zap:

index.js
```diff
@@ -1053,9 +1066,10 @@
                       constraints,
                       expArgTypes.concat([expRetType]),
                       values[idx],
-                      index,
+                      isNaN(index) ? expArgTypes.length : index,
```
davidchambers: How does this look, Kevin? It seems preferable to handle `NaN` sooner rather than later.

davidchambers: Actually, passing around `expTypes` rather than `expArgTypes` and `expRetType` turns out to be simpler. `NaN` is no longer necessary. :)

kedashoe: Nice change, looks good!

index.js
```diff
@@ -831,12 +839,17 @@
 
     var padding = _(showTypeSig_(expTypes.slice(0, index)));
 
+    //  f :: (String -> String) -> String
+    var f = function(f) {
+      return trimTrailingSpaces(underline(typePath[0], propPath, f));
+    };
+
```
davidchambers: While updating Sanctuary I noticed that I'd forgotten to apply `trimTrailingSpaces` below. I updated the definition of `f` accordingly. There's now no reason to curry `underline`, so I reverted that change.

index.js
```diff
@@ -158,7 +158,7 @@
 
   //  trimTrailingSpaces :: String -> String
   var trimTrailingSpaces = function(s) {
-    return s.replace(/[ ]+$/, '');
+    return s.replace(/[ ]+$/gm, '');
   };
```
davidchambers: Trimming all trailing spaces *after* formatting a message simplifies matters as we no longer need to clean up as we go.

index.js
```diff
@@ -1163,18 +1195,16 @@
...
                 throw invalidValue(name,
                                    constraints,
-                                   expArgTypes.concat([expRetType]),
+                                   expTypes,
                                    value,
                                    index,
                                    determineActualTypesLoose([value]));
               }
+              satisfactoryTypes(name, constraints, expTypes,
+                                $typeVarMap, value, index);
```
davidchambers: It turns out this function is only being called for side effects! At least now it's clear that this is the case.

index.js
davidchambers: #46

PR #47, opened 2016-04-03
underline nested type variable upon type-variable constraint violation

davidchambers:
This pull request should not be confused with #43, which has a similar title.

_Given `f :: [[a]] -> [[a]] -> [[a]]`, evaluate `f([['a', 'b', 'c'], [1, 2, 3]])`._

Before:

    Type-variable constraint violation

    concatNested :: Array (Array a) -> Array (Array a) -> Array (Array a)
                    ^^^^^^^^^^^^^^^
                           1

    1)  "a" :: String
        "b" :: String
        "c" :: String
        1 :: Number
        2 :: Number
        3 :: Number

    Since there is no type of which all the above values are members, the type-variable constraint has been violated.

After:

    Type-variable constraint violation

    concatNested :: Array (Array a) -> Array (Array a) -> Array (Array a)
                                 ^
                                 1

    1)  "a" :: String
        "b" :: String
        "c" :: String
        1 :: Number
        2 :: Number
        3 :: Number

    Since there is no type of which all the above values are members, the type-variable constraint has been violated.

### DIFF
--- a/index.js
+++ b/index.js
@@ -900,33 +900,54 @@
            underline(snd.typePath[0])(snd.propPath)(g);
   };
 
-  //  conflictingTypeVar :: ... -> Error
-  var conflictingTypeVar = function(
+  //  _typeVarConstraintViolation :: ... -> Error
+  var _typeVarConstraintViolation = function(
+    name,           // :: String
+    constraints,    // :: StrMap [TypeClass]
+    expTypes,       // :: [Type]
+    carets,         // :: String
+    numbers,        // :: String
+    pairss          // :: [[Pair Any [Type]]]
+  ) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1122,8 +1136,8 @@
               }
             } else {
               okTypes = determineActualTypesStrict(values);
-              if (isEmpty(okTypes)) {
-                throw conflictingTypeVar(
+              if (isEmpty(okTypes) && !isEmpty(values)) {
```
davidchambers: This check and the `!isEmpty(okTypes)` check below were removed—perhaps erroneously—in #38.

PR #51, opened 2016-04-10
fix type error in commonTypes

davidchambers:
Commit message:

> We currently assume that `commonTypes` is always applied to a list of lists of types of the same arity. If the first type is binary, for example, we incorrectly assume that all subsequent types also have `$1` and `$2` fields.

### DIFF
--- a/index.js
+++ b/index.js
@@ -139,6 +139,13 @@
     return result;
   };
 
+  //  reduce :: ([a], b, (b, a) -> b) -> b
+  var reduce = function(xs, y, f) {
+    var result = y;
+    for (var idx = 0; idx < xs.length; idx += 1) result = f(result, xs[idx]);
+    return result;
+  };
+
   //  strRepeat :: (String, Integer) -> String
   var strRepeat = function(s, times) {
     return Array(times + 1).join(s);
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -678,41 +681,22 @@
...
-    if (types[0].type === 'UNARY') {
-      $1s = {};
-      for (idx = 0; idx < types.length; idx += 1) {
-        t = types[idx];
-        if (t.$1.type !== 'UNKNOWN') $1s[t.$1.name] = t.$1;
-      }
-      candidates =
-      map(or(map(keys($1s), function(k) { return $1s[k]; }), [Unknown]),
-          UnaryType.from(types[0]));
-    } else if (types[0].type === 'BINARY') {
```
davidchambers: If the first type is binary…

index.js
```diff
@@ -678,41 +681,22 @@
...
-      $1s = {};
-      $2s = {};
-      for (idx = 0; idx < types.length; idx += 1) {
-        t = types[idx];
-        if (t.$1.type !== 'UNKNOWN') $1s[t.$1.name] = t.$1;
-        if (t.$2.type !== 'UNKNOWN') $2s[t.$2.name] = t.$2;
```
davidchambers: … we assume subsequent types also have `$1` and `$2` fields.

index.js
```diff
@@ -678,41 +681,22 @@
...
+  //  [[Array ???], [Array String], [Array ???]] ---> [Array String]
+  //
+  //  [[Either String ???], [Either ??? Number]] ---> [Either String Number]
+  var commonTypes = function($typeses, loose) {
+    return filter($typeses[0], function($t) {
+      return all($typeses, function(types) {
+        return any(types, function(t) {
+          var equal = equalTypes($t, t, loose);
+          if (equal && '$1' in $t && $t.$1.type === 'UNKNOWN') $t.$1 = t.$1;
+          if (equal && '$2' in $t && $t.$2.type === 'UNKNOWN') $t.$2 = t.$2;
```
davidchambers: Thanks to the mutation above we determine the type of `[Left('XXX'), Right(42)]` to be `Array (Either String Number)` rather than `Array (Either String ???)`.

:warning: Note that we're mutating `$typeses` in the process. I use `$` to draw attention to the mutation of an argument!

test/index.js
```diff
@@ -1002,7 +1002,7 @@ describe('def', function() {
                    '          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n' +
                    '                                                1\n' +
                    '\n' +
-                   '1)  {"end": 0, "start": 0} :: Object, StrMap Number, StrMap Number\n' +
+                   '1)  {"end": 0, "start": 0} :: Object, StrMap Number\n' +
```
davidchambers: The change incidentally fixes this problem too!

index.js
Avaq: Currently it's not only a mutation, it's a mutation of something shared (the input object). The caller seems to be aware of this (since it also prefixes the variable with `$`, but doesn't seem to benefit from, or use the mutation in any way. (Correct me if I'm wrong)

Wouldn't it be better to contain the mutation within the scope of `commonTypes`? That way The easiest way to do this would be deep-cloning the `$typeses[0]` object before you start filtering it. Alternatively, I think you can stay away from mutation altogether by using `reduce` for your iterations. If you want I can write the required reducers

---

I think you can do `tail` on `$typeses` before passing it to `all`. It would save you a single iteration, or in cases of a unary array, all of the work.

---

By using `any` to iterate the types, you're not actually collecting all the type info you *could*. For example if you have `[[Left String], [Left String, Right Number]]`, I think the output would be `[Either String ???]`(correct me if I'm wrong)  because `any` kept short-circuiting before you reached the `Right`. This too could be resolved with some custom `reduce`, not sure if you'd want to though.

davidchambers: Wow! This is one of the best comments I have received on a pull request. Thank you, Aldwin.

> The caller seems to be aware of this [mutation], but doesn't seem to benefit from [it] in any way. (Correct me if I'm wrong)

You're correct. At one point I even had `$typeses.length = 0;` to prevent us from relying on `$typeses` after applying `customTypes`.

> Alternatively, I think you can stay away from mutation altogether by using `reduce` for your iterations, and then you'd be able to return the improved type info without mutating the input. If you want I can write the required reducers, because I [already kinda did](https://gist.github.com/Avaq/b15b787bf837e6094e0ba1804a75c5e0) to lay out my thoughts.

Wonderful! I'll gladly use this approach.

> I think you can do `tail` on `$typeses` before passing it to `all`. It would save you a single iteration, or in cases of a unary array, all of the work.

I thought so too. We have `NaN`-like values, though, such as `Array Inconsistent` (the type of `[Left('XXX'), 42]`). When `commonTypes` is applied to `[[Array Inconsistent]]` the result should be `[]` when `loose` is `false` or `[Array Inconsistent]` when `loose` is `true` (for displaying `Array ???`).

> By using `any` to iterate the types, you're not actually collecting all the type info you *could*.

Great point! I'll be sure to include a test for this.

Avaq: > Wonderful! I'll gladly use this approach

I'm assuming the gist was enough to get you going. Or would you prefer if I work on the actual code?

> We have `NaN`-like values

Oh.. Yeah.. Of course we do. -,-

davidchambers:
I've incorporated your terrific [suggestion][1], @Avaq. Only minor changes were required in order for it to work. Most significantly, the initial value of `st.ok` (née `state.passed`) needs to be `true` rather than `false`. I made one stylistic change: replacing `.concat(x)` with `.concat([x])`. I like to pretend that `Array#concat` requires array arguments.


[1]: https://gist.github.com/Avaq/b15b787bf837e6094e0ba1804a75c5e0

Avaq:
> the initial value [...] needs to be true rather than false

I realized that whilst having a stroll just seconds ago, and I came running to the keyboard to let you know and found you had already spotted it. :)

index.js
davidchambers: > > By using `any` to iterate the types, you're not actually collecting all the type info you *could*.
>
> Great point! I'll be sure to include a test for this.

It's not clear to me *how* to test this. Since we've switched to an approach which does not have this problem, and since this pull request addresses a different issue, I'm not too concerned by the lack of a test case.

index.js
```diff
@@ -677,44 +684,46 @@
...
+    return reduce(typeses[0], [], function(types, t) {
+      var st = reduce(typeses, {ok: true, type: t}, function(st, types) {
+        var st$ = reduce(types, {ok: false, type: st.type}, function(st, t) {
+          var equal = equalTypes(st.type, t, loose);
+          return {ok: equal || st.ok,
+                  type: equal ? mergeTypes(st.type, t) : st.type};
         });
+        return {type: st$.type, ok: st.ok && st$.ok};
       });
+      return st.ok ? types.concat([st.type]) : types;
```
Avaq: I prefer `concat([x])` as well. It's not only stylistic in some cases because `concat(x)` does different things based on what `x` is, whereas `concat([x])` always behaves the same.

davidchambers: That's how I view it too. In *this* case it'd work the same either way, but I don't like to have to think unnecessarily. ;)

Avaq: > I don't like to have to think unnecessarily

I'm a programmer! Not a philosopher, for Pete's sake!

PR #55, opened 2016-04-13
underline invalid record value rather than entire record

davidchambers:
Alternative to #30

Commit message:

> When we determine that a value is not a member of the specified record type, we do not currently make the reason clear: we underline the type's entire string representation even if just one field is invalid.
>
> This problem compounds with nesting, since the outermost record type is underlined regardless of the depth at which validation fails.
>
> This commit updates the `underline` function to draw attention to the invalid field.
>
> Invalid `position.x`, before:
>
>     { color :: String, position :: { x :: Number, y :: Number } }
>     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>
> Invalid `position.x`, after:
>
>     { color :: String, position :: { x :: Number, y :: Number } }
>                                           ^^^^^^

Before:

```javascript
length({start: {x: 0, y: 0}, end: {x: null, y: null}});
// ! TypeError: Invalid value
//
//   length :: { end :: { x :: Number, y :: Number }, start :: { x :: Number, y :: Number } } -> Number
//             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
//                                                   1
//
//   1)  {"end": {"x": null, "y": null}, "start": {"x": 0, "y": 0}} :: Object, StrMap Object
//
//   The value at position 1 is not a member of ‘{ end :: { x :: Number, y :: Number }, start :: { x :: Number, y :: Number } }’.
```

After:

```javascript
length({start: {x: 0, y: 0}, end: {x: null, y: null}});
// ! TypeError: Invalid value
//
//   length :: { end :: { x :: Number, y :: Number }, start :: { x :: Number, y :: Number } } -> Number
//                             ^^^^^^
//                               1
//
//   1)  null :: Null
//
//   The value at position 1 is not a member of ‘Number’.
```

Do you think this would ease your pain, @rbuckheit?

### DIFF
--- a/index.js
+++ b/index.js
@@ -485,8 +485,14 @@
       }
       for (var idx = 0; idx < names.length; idx += 1) {
         var name = names[idx];
-        if (!has(name, x) || !fields[name]._test(x[name])) {
+        if (!has(name, x)) {
           return Left({value: x, typePath: [t], propPath: []});
+        }
+        var result = fields[name].validate(x[name]);
+        if (result.isLeft) {
+          return Left({value: result.value.value,
+                       typePath: [t].concat(result.value.typePath),
+                       propPath: [name].concat(result.value.propPath)});
         }
       }
       return Right(x);
@@ -1151,18 +1157,21 @@
         var t = type;
         var types = [t];
         for (var idx = 0; idx < propPath.length; idx += 1) {
-          types.push(t = t[propPath[idx]]);
+          types.push(t = (t.type === 'RECORD' ? t.fields : t)[propPath[idx]]);
         }
 
         var s = f(String(last(types)));
         for (idx = types.length - 2; idx >= 0; idx -= 1) {
+          var k = propPath[idx];
           t = types[idx];
           s = t.type === 'UNARY' ?
                 t.format(_, K(s)) :
-              t.type === 'BINARY' && propPath[idx] === '$1' ?
+              t.type === 'BINARY' && k === '$1' ?
                 t.format(_, K(s), _) :
+              t.type === 'BINARY' && k === '$2' ?
+                t.format(_, _, K(s)) :
               // else
-                t.format(_, _, K(s));
+                t.format(_, function(k$) { return k$ === k ? K(s) : _; });  // jshint ignore:line
         }
 
         return isParameterizedType(type) ? s.slice(1, -1) : s;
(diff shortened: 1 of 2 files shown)


soroushjp:
@davidchambers Looks good to me overall. I'll be honest, I'm struggling to follow the implementation too closely, but the test seems sound to me.

Smaller usability comment, but `position <n>` is a little ambiguous to me. It could refer to the first argument to the function or the first key (alphabetically or as declared) in the record object.

davidchambers:
> Smaller usability comment, but `position <n>` is a little ambiguous to me.

    f :: Foo -> { bar :: Bar } -> Baz
                         ^^^
                          1

`position <n>` refers to the underlined portion of the type signature labelled `<n>`. I'm not sure how to make this clearer. It certainly doesn't work well when the type signature is so long it must be wrapped.

kedashoe:
Maybe adding "underlined" would help a bit? `The underlined value at position <n>...` or `The value underlined above at position <n>...`

davidchambers:
Great idea, Kevin! Let's do that. I'll open a separate pull request.

davidchambers:
Hmm. I just realized that the underlined text represents a type rather than a value. So it's an __underlined *type*__ rather than an __underlined value__.

We could write `The value at position <n> (underlined)...`, but I'm not sure this adds clarity (and it's actually slightly confusing in cases where multiple types are underlined).

kedashoe:
> I just realized that the underlined text represents a type rather than a value. So it's an underlined type rather than an underlined value.

hmm.. I think in my mind "value" here was just a generic word for what is being pointed to, ie "The **thing** at position 1...". But I can see how this might be confusing and no generic words/phrases are immediately coming to mind that would sound great here.

I don't have the repo on this computer, but a quick look at the source.. it might be pretty easy to know what we are underlining? Could we do something like

- `invalidValue`: `The underlined value..`
- `typeVarConstraintViolation`: `The underlined type..`
- `typeVarConstraintViolation2`: `The underlined type..`
- `typeClassConstraintViolation`: `The underlined constraint..`

davidchambers:
> Could we do something like
>
>   - `invalidValue`: `The underlined value..`

Even in the case of an invalid value we underline (the string representation of) a type.

Avaq:
How about "the value *provided for* position `n`"?

davidchambers:
I've rebased this branch. After we merge #58 this diff will be quite small.

Issue #56, opened 2016-04-13
‘test’ methods of unary and binary types no longer check nested values


davidchambers:
Much has changed in this project in recent weeks. With #48 we moved the logic for testing nested values from the `test` methods of unary and binary types to a central `testNested` function. This broke a useful but currently untested feature: the ability to use the `test` method of a unary or binary type directly, as in `$.Array($.Number).test(x)` or `Pair(a, a).test(x)`.

To support type variables it's necessary to have access to the environment. We could provide:

```haskell
$.test :: [Type] -> Type -> Any -> Boolean
```

It could be used like so:

```javascript
const $ = require('sanctuary-def');

const Direction = require('./types/Direction');
const Pair = require('./types/Pair');

//    env :: [Type]
const env = $.env.concat([Direction, Pair]);

//    def :: (String, StrMap [TypeClass], [Type], Function) -> Function
const def = $.create(true, env);

//    is :: Type -> Any -> Boolean
const is = $.test(env);

//    a :: Type
const a = $.TypeVariable('a');

is($.Array($.Number), [1, 2, 3]);  // => true
is(Pair(a, a), ['foo', 'bar']);  // => true
is(Pair(a, a), ['foo', 123]);  // => false
```

It would be good to prevent people from using the `test` methods directly, as they cannot support type variables. One possibility is to rename the field `_test` (to suggest that it's for internal use), and to remove mentions of the field from the documentation.

Whatever we do we must add tests!

PR #57, opened 2016-04-19
make satisfactoryTypes pure to permit reuse

davidchambers:
This is a step towards addressing #56. I'd like to reuse `satisfactoryTypes` in the proposed `$.test` function, but it would currently require `try`–`catch`, which is clearly unacceptable! ;)

### DIFF
--- a/index.js
+++ b/index.js
@@ -29,6 +29,30 @@
   var hasOwnProperty    = Object.prototype.hasOwnProperty;
   var toString          = Object.prototype.toString;
 
+  //  Left :: a -> Either a b
+  var Left = function(x) {
+    return {
+      '@@type': 'sanctuary-def/Either',
+      isLeft: true,
+      isRight: false,
+      chain: function(f) { return this; },
+      map: function(f) { return this; },
+      value: x
+    };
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -29,6 +29,30 @@
...
+  var Right = function(x) {
+    return {
+      '@@type': 'sanctuary-def/Either',
+      isLeft: false,
+      isRight: true,
+      chain: function(f) { return f(this.value); },
+      map: function(f) { return Right(f(this.value)); },
+      value: x
+    };
+  };
```
soroushjp: Amazing how simple `Either` implementations are at the end of the day.

soroushjp:
LGTM. Would it make sense to write tests for `_satisfactoryTypes` this function as we change its implementation?

index.js
davidchambers: Yep! Simple, principled things make the only reasonable foundation for complex programs. :)

svozza:
Looks good. I'm excited about `$.test`.

davidchambers:
Superseded by #58

PR #58, opened 2016-04-20
define $.test

### LINKED ISSUE #56: ‘test’ methods of unary and binary types no longer check nested values

davidchambers:
Fixes #56

Commit message:

> Currently, one determines whether a value is a member of a type via:
>
>     t.test(x)
>
> This commit introduces a public function for performing such checks:
>
>     $.test(env, t, x)
>
> This function takes an environment in order to handle type variables.

I'll comment on the interesting sections of the diff in the morning. :zzz:

### DIFF
--- a/index.js
+++ b/index.js
@@ -29,6 +29,30 @@
   var hasOwnProperty    = Object.prototype.hasOwnProperty;
   var toString          = Object.prototype.toString;
 
+  //  Left :: a -> Either a b
+  var Left = function Left(x) {
+    return {
+      '@@type': 'sanctuary-def/Either',
+      isLeft: true,
+      isRight: false,
+      chain: function(f) { return Left(x); },
+      map: function(f) { return Left(x); },
+      value: x
+    };
+  };
+
+  //  Right :: b -> Either a b
+  var Right = function Right(x) {
+    return {
+      '@@type': 'sanctuary-def/Either',
+      isLeft: false,
+      isRight: true,
+      chain: function(f) { return f(x); },
+      map: function(f) { return Right(f(x)); },
+      value: x
+    };
+  };
+
   //  K :: a -> b -> a
   var K = function(x) { return function(y) { return x; }; };
 
@@ -42,6 +66,16 @@
 
   //  always :: a -> (-> a)
   var always = function(x) { return function() { return x; }; };
+
+  //  assoc :: (String, a, StrMap a) -> StrMap a
+  var assoc = function(key, value, strMap) {
+    var result = {};
+    for (var k in strMap) {
+      result[k] = strMap[k];
+    }
+    result[key] = value;
+    return result;
+  };
 
   //  chain :: ([a], (a -> [b])) -> [b]
   var chain = function(xs, f) {
@@ -232,8 +266,15 @@
   $.TypeClass = function(name, test) {
     return {
       name: name,
-      test: test,
+      _test: test,
       toString: always(stripNamespace(name))
+    };
+  };
+
+  //  testFrom :: (a -> Result) -> a -> Boolean
+  var testFrom = function(validate) {
+    return function(x) {
+      return validate(x).isRight;
     };
   };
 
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -232,24 +266,31 @@
...
-    test: K(true),
+    validate: Right,
+    _test: K(true),
     toString: always('???')
   };
 
   //  Inconsistent :: Type
   var Inconsistent = {
-    '@@type': 'sanctuary-def/Inconsistent',
+    '@@type': 'sanctuary-def/Type',
```
davidchambers: The old value was erroneous.

davidchambers:
I was able to avoid the incidental API change I'd planned to explain today, so this pull request is now ready to be reviewed. :)

index.js
```diff
@@ -307,24 +372,44 @@
...
+        if (!test(x)) {
+          return Left({value: x, typePath: [t], propPath: []});
+        }
+        for (var n = 1; n <= 2; n += 1) {
+          var _ = '_' + String(n);
+          var $ = '$' + String(n);
+          for (var idx = 0, xs = t[_](x); idx < xs.length; idx += 1) {
+            var result = t[$].validate(xs[idx]);
+            if (result.isLeft) {
+              return Left({value: result.value.value,
```
svozza: `value.value`?

index.js
```diff
@@ -693,40 +795,253 @@
...
+    if (isEmpty(values)) return [Unknown];
+    //  typeses :: [[Type]]
+    var typeses = map(values, function(value) {
+      var seen$;
+      if (typeof value === 'object' && value != null ||
+          typeof value === 'function') {
+        //  Abort if a circular reference is encountered; add the current
+        //  object to the list of seen objects otherwise.
+        if (seen.indexOf(value) >= 0) return [];
+        seen$ = seen.concat([value]);
```
svozza: Do we really have to clone the array every time here? This function is going to be invoked a lot, it might make sense to mutate.

index.js
davidchambers:
```haskell
result :: Either { value :: Any, typePath: [Type], propPath: [String] } Any
```

The first `.value` provides access to the Left's value, which happens to be a record with a `value` field. ;)

index.js
davidchambers: > Do we really have to clone the array every time here?

I believe we do. See #53.

test/index.js
```diff
@@ -2003,3 +2058,42 @@ describe('def', function() {
   });
 
 });
+
+describe('test', function() {
+
```
svozza: Is there any value in using types other than `$.Number`?

index.js
svozza: Ah sorry, I missed that one. It's a pity we don't have a cons operator and could just do `seen : value`.

test/index.js
davidchambers: We still have the tests you added for many of the exported types. :)

This block is intended to test `$.test`, particularly in relation to:

  - concrete parameterized types such as `$.Array($.Number)`, for which `t.test(['a', 'b', 'c'])` has erroneously evaluated `true` since #48; and

  - polymorphic parameterized types such as `$.Array(a)`, for which `t.test([1, 2, '3'])` has always erroneously evaluated `true`.

index.js
davidchambers: `value : seen` would be very nice indeed.

test/index.js
svozza: Yep, makes sense.

svozza:
I can't wait to start using this function.

PR #65, opened 2016-05-07
update $.create to take a record rather than positional arguments

davidchambers:
@kedashoe suggested in sanctuary-js/sanctuary#206 that we avoid Boolean arguments. I agree that we should do so. If we're to have `S.create` take an options record, we should update `$.create` to match.


svozza:
Nice. Looks good.

PR #68, opened 2016-05-09
types: define Pair type constructor

davidchambers:
Pair is an important type constructor. Without it, `R.toPairs({foo: 42})` evaluates to an "invalid" value. See sanctuary-js/sanctuary#214 for discussion of `S.pairs`, which could make use of Pair.


Avaq:
It looks perfectly fine to me. :)

gilligan:
Looks good to me, Having a `Pair` type makes sense to me.

svozza:
Nice. I think I will hold off on `S.toPairs` until this type lands.

PR #71, opened 2016-05-15
Run tests in parallel

svozza:
The Circle CI docs seem to be out of date so this is a test to see if they support Node 6 yet. I've also replicated the parallel build we have in Sanctuary.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,17 +1,15 @@
 dependencies:
   override:
-    - make setup
+    - rm -rf node_modules
+    - case $CIRCLE_NODE_INDEX in 0) make setup ;; 1) nvm exec 0.12.7 make setup ;; 2) nvm exec 4 make setup ;; 3) nvm exec 5 make setup ;; esac:
+        parallel: true
 
 machine:
   node:
-    version: 0.10.34
+    version: 0.10.45
 
 test:
   override:
     - make lint
-    - make test
-    - nvm install 0.12.7
-    - nvm alias default 0.12.7
-    - make test
-    - nvm use 4 && make test
-    - nvm use 5 && make test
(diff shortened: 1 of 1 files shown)


svozza:
Does this actually run the build in parallel? The output on CircleCI only shows one test run.

davidchambers:
> Does this actually run the build in parallel?

One must opt into parallelism in the CircleCI UI. I have made this change (to 4x, the highest free setting).

svozza:
Ah yes, I did have a dig around the parallel settings in there but they kept trying to charge money so I just left it be.

circle.yml
```diff
@@ -1,17 +1,15 @@
 dependencies:
   override:
-    - make setup
+    - rm -rf node_modules
+    - case $CIRCLE_NODE_INDEX in 0) make setup ;; 1) nvm exec 0.12.7 make setup ;; 2) nvm exec 4 make setup ;; 3) nvm exec 5 make setup ;; 4) nvm exec 6 make setup ;; esac:
+        parallel: true
```
davidchambers: Since parallelism is limited to 4x, let's drop one of the older Node versions. Having Node 0.10.x in the mix is useful, as it's a reasonable approximation of old browsers. CircleCI was able to flag sanctuary-js/sanctuary-type-classes#2 for its dependence on [`Array.of`][1], which is not defined in Node 0.10.x. Shall we drop Node 0.12.7?


[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/of

svozza: According to the Node.js Foundation [user survey](https://nodejs.org/static/documents/2016-survey-report.pdf) only 18% are on Node 0.12 so I'm happy with dropping it.

svozza:
Ah so it looks like 6 is not in yet.

davidchambers:
Would you mind rebasing this branch, Stefano?

svozza:
Sure no problem!

davidchambers:
Thanks!

PR #79, opened 2016-06-26
define $.Function and $.UnaryTypeVariable

davidchambers:
__TL;DR__ we can now do this:

```javascript
const b = $.TypeVariable('b');
const f = $.UnaryTypeVariable('f');

//    Functor :: TypeClass
const Functor = ...;

//    map :: Functor f => (a -> b) -> f a -> f b
const map =
def('map',
    {f: [Functor]},
    [$.Function([a, b]), f(a), f(b)],
    (fn, functor) => functor.map(fn));
```

---

_This pull request sits atop #78 and should not be merged until #78 has been merged._

_sanctuary-js/sanctuary#232 is heavily dependent on this pull request._

This pull request adds support for what I believe to be the two missing pieces of the puzzle:

  - higher-order functions; and
  - parameterized type variables.

A wonderful consequence of this change is that type signatures in error messages will now match the type signatures which appear in documentation – no more `map :: Function -> a -> b`!

#### `$.Function :: Array Type -> Type`

The type formerly known as `$.Function` is now known as `$.AnyFunction`. `$.Function` is now a *constructor* for function types.

Examples:

  - `$.Function([$.Date, $.String])` represents the `Date -> String` type; and
  - `$.Function([a, b, a])` represents the `(a, b) -> a` type.

```javascript
//    formatDates :: (Date -> String) -> Array Date -> String
const formatDates =
def('formatDates',
    {},
    [$.Function([$.Date, $.String]), $.Array($.Date), $.String],
    (f, dates) => implementation);
```

What will happen when one evaluates `formatDates(d => d.toISOString(), [new Date(0)])`?

1.  A check will be performed to confirm that `d => d.toISOString()` is in fact a function (of some sort).

2.  A check will be performed to confirm that `[new Date(0)]` is in fact an array.

3.  A check will be performed to confirm that `new Date(0)` is in fact a Date object.

4.  `f` will be decorated along these lines:

    ```javascript
    function(x) {
      if (arguments.length !== 1) throw new TypeError('...');
      if (!$.Date._test(x)) throw new TypeError('...');
      var output = f(x);
      if (!$.String._test(output)) throw new TypeError('...');
      return output;
    }
    ```

5.  `implementation` will be applied to the above function and `dates`.

`$.Function` takes an array of types just as `def` does. `$.Function` can thus be used to represent functions of any arity (including nullary functions). The downside of this approach is that representing curried functions is awkward: `a -> b -> a`, for example, is `$.Function([a, $.Function([b, a])])`. The solution is to define a binary type constructor:

```javascript
//    Fn :: (Type, Type) -> Type
const Fn = (x, y) => $.Function([x, y]);
```

One can then express `a -> b -> a` as `Fn(a, Fn(b, a))`.

#### `$.UnaryTypeVariable :: String -> Type -> Type`

This essentially combines [`$.TypeVariable`][1] and [`$.UnaryType`][2]. See the readme for details.

#### Limitations

  - Only functions at the top level of the types array are wrapped.

    The function in `[$.Array($.Function([a, b])), $.Array(a), $.Array(b)]`, for example, will not be wrapped. I hope to address this limitation in the future.

#### Breaking API changes

  - `$.Function` has been repurposed. When upgrading the most straightforward substitution is `:%s/\$\.\<Function\>/$.AnyFunction/g`, but in most cases one will wish to faithfully describe the types of one's higher-order functions (using the new `$.Function`).

  - “Extractors” of parameterized types are no longer exposed via `._1` and `._2`. They are now exposed via `.types.$1.extractor` and `.types.$2.extractor`.

  - The inner types of parameterized types are no longer exposed via `.$1` and `.$2`. They are now exposed via `.types.$1.type` and `.types.$2.type`.

  - RecordType values no longer expose `.fields`. The type of the `x` field, for example, was previously exposed via `.fields.x`. It is now exposed via `.types.x.type`. The similarity between RecordType values and values of parameterized types is intentional, and allows code sharing where previously branching logic was required. If one squints one can see a type such as `Pair a b` as a record type: `{ $1 :: a, $2 :: b }`.

#### Style changes

  - In accordance with sanctuary-js/sanctuary#180 `[a]` is now written `Array a`.

  - In accordance with sanctuary-js/sanctuary#214 an ES5-compatible environment is now assumed, so the implementation makes use of `Array#map`, `Object.keys`, and other useful ES5 functions.

  - In accordance with sanctuary-js/sanctuary#225 the test suite now defines `Nothing :: Maybe a` rather than `Nothing :: -> Maybe a`.

#### Refactoring

It took many, many hours but I was successful in creating a function, `createType`, which makes it clear what is required of a type and removes duplication.

`$.TypeVariable`, before:

```javascript
//  TypeVariable :: String -> Type
$.TypeVariable = function(name) {
  return {
    '@@type': 'sanctuary-def/Type',
    type: 'VARIABLE',
    name: name,
    validate: Right,
    _test: K(true),
    toString: always(name)
  };
};
```

`$.TypeVariable`, after:

```javascript
//  TypeVariable :: String -> Type
$.TypeVariable = function(name) {
  return createType(VARIABLE, name, always2(name), K(true), [], {});
};
```

`$.UnaryType`, before:

```javascript
//  UnaryType :: (String, (x -> Boolean), (t a -> [a])) -> Type -> Type
var UnaryType = $.UnaryType = function(name, test, _1) {
  return function($1) {
    var format = function(f, f$1) {
      return f('(' + stripNamespace(name) + ' ') + f$1(String($1)) + f(')');
    };
    var validate = function(x) {
      if (!test(x)) {
        return Left({value: x, typePath: [t], propPath: []});
      }
      for (var idx = 0, xs = _1(x); idx < xs.length; idx += 1) {
        var result = $1.validate(xs[idx]);
        if (result.isLeft) {
          return Left({value: result.value.value,
                       typePath: [t].concat(result.value.typePath),
                       propPath: ['$1'].concat(result.value.propPath)});
        }
      }
      return Right(x);
    };
    var t = {
      '@@type': 'sanctuary-def/Type',
      type: 'UNARY',
      name: name,
      validate: validate,
      _test: testFrom(validate),
      format: format,
      toString: always(format(id, id)),
      _1: _1,
      $1: $1
    };
    return t;
  };
};
```

`$.UnaryType`, after:

```javascript
//  UnaryType :: (String, (x -> Boolean), (t a -> Array a)) -> Type -> Type
var UnaryType = $.UnaryType = function(name, test, _1) {
  return function($1) {
    var format = function(outer, inner) {
      return outer('(' + stripNamespace(name) + ' ') +
             inner('$1')(String($1)) + outer(')');
    };
    var types = {$1: {extractor: _1, type: $1}};
    return createType(UNARY, name, format, test, ['$1'], types);
  };
};
```

`$.Function` and `$.UnaryTypeVariable` are also defined via `createType`.

---

This is a great opportunity to familiarize yourself with this codebase if you're interested but haven't known where to begin. The code should be easier to follow now, though there are still several functions which take six, seven, or even nine (!) arguments without many comments beyond the type annotations. I'm happy to answer questions; please let me know if anything is unclear. Future readers (including future me) will appreciate anything we can do now to clarify the code.

[1]: https://github.com/sanctuary-js/sanctuary-def#typevariable
[2]: https://github.com/sanctuary-js/sanctuary-def#unarytype

### DIFF
--- a/index.js
+++ b/index.js
@@ -20,10 +20,10 @@
   var MAX_SAFE_INTEGER = Math.pow(2, 53) - 1;
   var MIN_SAFE_INTEGER = -MAX_SAFE_INTEGER;
 
-  var LEFT_SINGLE_QUOTATION_MARK = '\u2018';
-  var RIGHT_SINGLE_QUOTATION_MARK = '\u2019';
-
+  var forEach           = Array.prototype.forEach;
+  var map               = Array.prototype.map;
   var push              = Array.prototype.push;
+  var slice             = Array.prototype.slice;
   var hasOwnProperty    = Object.prototype.hasOwnProperty;
   var toString          = Object.prototype.toString;
 
@@ -54,82 +54,41 @@
   //  K :: a -> b -> a
   var K = function(x) { return function(y) { return x; }; };
 
-  //  all :: ([a], (a -> Boolean)) -> Boolean
-  var all = function(xs, pred) {
-    for (var idx = 0; idx < xs.length; idx += 1) {
-      if (!pred(xs[idx])) return false;
-    }
-    return true;
-  };
-
   //  always :: a -> (-> a)
   var always = function(x) { return function() { return x; }; };
 
-  //  assoc :: (String, a, StrMap a) -> StrMap a
-  var assoc = function(key, value, strMap) {
-    var result = {};
-    for (var k in strMap) {
-      result[k] = strMap[k];
-    }
-    result[key] = value;
-    return result;
-  };
+  //  always2 :: a -> (b, c) -> a
+  var always2 = function(x) { return function(y, z) { return x; }; };
 
-  //  chain :: ([a], (a -> [b])) -> [b]
+  //  chain :: (Array a, a -> Array b) -> Array b
   var chain = function(xs, f) {
     var result = [];
-    for (var idx = 0; idx < xs.length; idx += 1) {
-      push.apply(result, f(xs[idx]));
-    }
+    xs.forEach(function(x) { push.apply(result, f(x)); });
     return result;
   };
-
-  //  filter :: ([a], (a -> Boolean)) -> [a]
-  var filter = function(xs, pred) {
-    var result = [];
-    for (var idx = 0; idx < xs.length; idx += 1) {
-      if (pred(xs[idx])) {
-        result.push(xs[idx]);
-      }
-    }
-    return result;
-  };
-
-  //  has :: (String, Object) -> Boolean
-  var has = function(key, obj) { return hasOwnProperty.call(obj, key); };
 
   //  id :: a -> a
   var id = function(x) { return x; };
 
-  //  isEmpty :: [a] -> Boolean
+  //  isEmpty :: Array a -> Boolean
   var isEmpty = function(xs) { return xs.length === 0; };
 
-  //  keys :: Object -> [String]
-  var keys = function(obj) {
-    var result = [];
-    for (var key in obj) if (has(key, obj)) result.push(key);
-    return result.sort();
-  };
-
-  //  last :: [a] -> a
-  var last = function(xs) { return xs[xs.length - 1]; };
-
-  //  map :: ([a], (a -> b)) -> [b]
-  var map = function(xs, f) {
-    var result = [];
-    for (var idx = 0; idx < xs.length; idx += 1) result.push(f(xs[idx]));
-    return result;
-  };
-
-  //  or :: ([a], [a]) -> [a]
-  var or = function(xs, ys) { return isEmpty(xs) ? ys : xs; };
-
-  //  prefix :: String -> String -> String
-  var prefix = function(x) {
-    return function(y) {
-      return x + y;
+  //  isPrefix :: Array a -> Array a -> Boolean
+  var isPrefix = function(candidate) {
+    return function(xs) {
+      if (candidate.length > xs.length) return false;
+      for (var idx = 0; idx < candidate.length; idx += 1) {
+        if (candidate[idx] !== xs[idx]) return false;
+      }
+      return true;
     };
   };
+
+  //  last :: Array a -> a
+  var last = function(xs) { return xs[xs.length - 1]; };
+
+  //  or :: (Array a, Array a) -> Array a
+  var or = function(xs, ys) { return isEmpty(xs) ? ys : xs; };
 
   //  quote :: String -> String
   var quote = function(s) {
@@ -147,17 +106,10 @@
     return '"' + escaped.replace(/"/g, '\\"') + '"';
   };
 
-  //  range :: (Number, Number) -> [Number]
+  //  range :: (Number, Number) -> Array Number
   var range = function(start, stop) {
     var result = [];
     for (var n = start; n < stop; n += 1) result.push(n);
-    return result;
-  };
-
-  //  reduce :: ([a], b, (b, a) -> b) -> b
-  var reduce = function(xs, y, f) {
-    var result = y;
-    for (var idx = 0; idx < xs.length; idx += 1) result = f(result, xs[idx]);
     return result;
   };
 
@@ -176,9 +128,9 @@
   //  _ :: String -> String
   var _ = r(' ');
 
-  //  toPairs :: StrMap a -> [Pair String a]
-  var toPairs = function(obj) {
-    return map(keys(obj), function(k) { return [k, obj[k]]; });
+  //  stripOutermostParens :: String -> String
+  var stripOutermostParens = function(s) {
+    return s.slice('('.length, -')'.length);
   };
 
   //  trimTrailingSpaces :: String -> String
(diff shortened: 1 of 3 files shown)


svozza:
This looks great, David! I should have a chance to dig into the code tomorrow evening.

index.js
```diff
@@ -219,41 +164,43 @@
       };
     };
 
-    switch (typeOf(x)) {
```
svozza: Any particular reason to remove `typeOf`?

index.js
```diff
@@ -804,234 +740,414 @@
...
+    values          // :: Array Any
+  ) {
+    var $typeVarInfo = {};
+    for (var typeVarName in typeVarInfo) {
+      var info = typeVarInfo[typeVarName];
+      var $values = {};
+      for (var key in info.values) $values[key] = info.values[key].slice();
+      $typeVarInfo[typeVarName] = {types: info.types.slice(), values: $values};
+    }
+    if (!hasOwnProperty.call($typeVarInfo, typeVar.name)) {
```
svozza: If you use `Object.create(null)` to create `$typeVarInfo`, you should be able to avoid `hasOwnProperty`.

index.js
davidchambers: It was part of why quest to reduce the number of helper functions. It was only used in two places so I decided it wasn't pulling its weight. `toString.call(x)` has the advantage of being more explicit too, once one is aware that `toString` is a reference to `Object.prototype.toString`.

I'm happy to reinstate `typeOf` if you think it improves readability.

index.js
davidchambers: Good point! It had not occurred to me that `Object.create` is now at our disposal.

We'd still need a membership check of some description. `key in obj` is certainly more pleasant to read than `hasOwnProperty.call(obj, key)`, but I'm not sure it warrants the mental overhead of having `Object.create(null)` in some places and `{}` in others. Or would you suggest `Object.create(null)` on lines 782, 784, and 787 as well?

index.js
svozza: No, not at all I think you're right to make things more explicit.

index.js
svozza: Yeah, I didn't want to leave the same comment on multiple line so I meant using it across the board in this function. `Object.create(null)` makes it explicit that these objects are to be used purely as associative arrays and not for any other purpose.

davidchambers: I realized that we need only use `Object.create(null)` for string maps (not for records). I've made the change you suggested. :)

index.js
```diff
@@ -264,250 +212,256 @@
...
+    var validate = function(x) {
+      if (!test(x)) return Left({value: x, propPath: []});
+      for (var idx = 0; idx < keys.length; idx += 1) {
+        var k = keys[idx];
+        var t = types[k];
+        for (var idx2 = 0, ys = t.extractor(x); idx2 < ys.length; idx2 += 1) {
+          var result = t.type.validate(ys[idx2]);
+          if (result.isLeft) {
+            var value = result.value.value;
+            var propPath = [k].concat(result.value.propPath);
```
svozza: Is there a reason not to use local mutation here, I seem to remember some issue from a while back.

davidchambers: You may be thinking of #53. I *think* it would be okay to use local mutation here, but it would take significant mental effort to convince myself that no one else is holding a reference to the array. I don't see a significant downside to creating a new array here, particularly since we're about to short-circuit.

test/index.js
```diff
@@ -2225,6 +2416,103 @@ describe('def', function() {
...
+           errorEq(TypeError,
+                   'Invalid value\n' +
+                   '\n' +
+                   'sum :: Foldable f => f FiniteNumber -> FiniteNumber\n' +
+                   '                       ^^^^^^^^^^^^\n' +
+                   '                            1\n' +
+                   '\n' +
+                   '1)  "foo" :: String\n' +
+                   '\n' +
+                   'The value at position 1 is not a member of ‘FiniteNumber’.\n'));
```
davidchambers: As a result of sanctuary-js/sanctuary#234 I realized we lacked tests for the two cases above. After writing the tests it was apparent that we also lacked logic to handle these two cases. I've made some changes so these tests now pass.

/cc @jdudek

index.js
svozza: Yep, best leave it as it is.

index.js
davidchambers: I changed my mind. I found myself reluctant to replace `hasOwnProperty.call(o, k)` with `k in o` in various places because I was unsure of whether `o` was created via `Object.create(null)`. Without a type system it would be all too easy for a `{}` to slip through, and expecting contributors to mentally trace the ancestry of an object in order to determine the appropriate membership check is asking too much.

Although using `Object.create(null)` is "correct", ignoring inherited properties when performing membership checks is less likely to result in bugs than "promising" never to create objects with inherited properties.

I do wonder whether we should rewrite sanctuary-def in PureScript one day. ;)

svozza: Yeah, fair enough. I guess the real issue is that we should be using the native `Map` type for this stuff but we can't.

PR #82, opened 2016-09-26
provide a placeholder value

davidchambers:
With sanctuary-js/sanctuary#216, Sanctuary is becoming more an alternative to Ramda rather than a complement. It doesn't make sense to continue to rely on Ramda to provide the placeholder value. I found myself defining `S.__` but then realized that the placeholder should be defined here since it is sanctuary-def which actually makes use of it. It's probably still a good idea to define `S.__`, but we can simply make it an :alien: once `$.__` is defined.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,19 +86,16 @@ inc(7);
 One may wish to partially apply a function whose parameters are in the "wrong"
 order. All functions defined via sanctuary-def accommodate this by accepting
 "placeholders". A placeholder is an object with a `'@@functional/placeholder'`
-property whose value is `true`. [`R.__`][R.__] is one such object.
-A placeholder indicates an argument yet to be provided. For example:
+property whose value is `true`. `$.__` is one such object. A placeholder
+indicates an argument yet to be provided. For example:
 
 ```javascript
-//    _ :: Placeholder
-const _ = {'@@functional/placeholder': true};
-
 //    concatS :: String -> String -> String
 const concatS =
 def('concatS', {}, [$.String, $.String, $.String], (x, y) => x + y);
 
 //    exclaim :: String -> String
-const exclaim = concatS(_, '!');
+const exclaim = concatS($.__, '!');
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1029,14 +1026,13 @@ Multiple constraints may be placed on a type variable by including multiple
...
-[3]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SyntaxError
-[4]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError
-[5]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MIN_SAFE_INTEGER
-[6]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER
-[7]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create
-[8]: https://en.wikipedia.org/wiki/Enumerated_type
-[9]: http://ramdajs.com/docs/#toString
-[10]: https://en.wikipedia.org/wiki/Semigroup
-[11]: https://github.com/fantasyland/fantasy-land#semigroup
+[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments
```
safareli: Maybe change them to some non number (NaN :D) values so we don't need to update all of them when one link is removed/added, as in sanctuary-type-classes

davidchambers: Good idea. I'll do so now in a separate pull request.

davidchambers: #84

safareli: Sweet!

davidchambers: I rebased this branch after merging #84. The changes to __README.md__ are much clearer now. Would you mind having another look?

safareli: Nice! looks good

PR #83, opened 2016-09-26
sanctuary-type-classes@1.0.x

davidchambers:
This is the most important change:

``` diff
   //  TypeClass :: Type
-  var TypeClass = type0('sanctuary-def/TypeClass');
+  var TypeClass = type0('sanctuary-type-classes/TypeClass');
```

This is necessary to make sanctuary-def compatible with TypeClass values created via `Z.TypeClass` from sanctuary-type-classes.

The most satisfying change is the removal of `show`; we can now use `Z.toString`. Also satisfying is the removal of a comment explaining the use of `show` to determine equality in `$.EnumType`. Thanks to `Z.equals` we're no longer relying on string representations.

Ramda is no longer a dev dependency, since we can now use `Z.toString` in place of `R.toString` in `eq`, our equivalence assertion function.

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,14 +4,14 @@
 
   /* istanbul ignore else */
   if (typeof module === 'object' && typeof module.exports === 'object') {
-    module.exports = f();
+    module.exports = f(require('sanctuary-type-classes'));
   } else if (typeof define === 'function' && define.amd != null) {
-    define([], f);
+    define(['sanctuary-type-classes'], f);
   } else {
-    self.sanctuaryDef = f();
+    self.sanctuaryDef = f(self.sanctuaryTypeClasses);
   }
 
-}(function() {
+}(function(Z) {
 
   'use strict';
 
@@ -21,8 +21,6 @@
   var MIN_SAFE_INTEGER = -MAX_SAFE_INTEGER;
 
   var forEach           = Array.prototype.forEach;
-  var map               = Array.prototype.map;
-  var push              = Array.prototype.push;
   var slice             = Array.prototype.slice;
   var hasOwnProperty    = Object.prototype.hasOwnProperty;
   var toString          = Object.prototype.toString;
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -10,13 +10,14 @@
   "scripts": {
     "test": "make test lint"
   },
-  "dependencies": {},
+  "dependencies": {
+    "sanctuary-type-classes": "sanctuary-js/sanctuary-type-classes#dc-everything"
```
davidchambers: :warning: Update before merging.

davidchambers:
I believe this pull request is now ready to merge. I would :heart: a review.

index.js
```diff
@@ -359,10 +268,10 @@
   };
 
   //  $$type :: a -> String
-  var $$type = function(x) {
-    return x != null && toString.call(x['@@type']) === '[object String]' ?
+  var $$type = function $$type(x) {
+    return x != null && $$type(x['@@type']) === 'String' ?
```
Avaq: `$$type({'@@type': {'@@type': 'String'}}) === {'@@type': 'String'}` which invalidates the type signature. Not sure if it really matters though.

test/index.js
```diff
@@ -1831,7 +1840,16 @@ describe('def', function() {
...
+    def('unnest',
+        {},
+        [$.Array($.Array(a)), $.Array(a)],
+        function(xss) {
+          var xs = [];
+          for (var idx = 0; idx < xss.length; idx += 1) {
+            Array.prototype.push.apply(xs, xss[idx]);
+          }
+          return xs;
+        });
```
Avaq: Could you just do `return Z.chain(identity, xss)`?

index.js
davidchambers: Haha! Good point. I reverted this change.

test/index.js
davidchambers: Indeed! This led me to realize that several lambdas in __test/index.js__ could be replaced with `Z.concat`, `Z.map`, or `Z.filterM`. This is nice as it keeps the tests focused on the *types* of the functions.

davidchambers:
Would you mind taking one more :eyes:, @Avaq?

Avaq:
It looks good to me, but I must admit I'm a bit lost in some of the diffs. They would require more of my time to truly understand, and I'm not sure if it's worth it because they appear to be simple refactors.

PR #86, opened 2016-10-08
define ‘toString’ for functions returned by ‘def’

davidchambers:
Sanctuary's test suite is testing too many things. Initially the currying and type-checking code lived in Sanctuary, so it was important to test this code thoroughly in Sanctuary's test suite. Now this code lives in sanctuary-def, and sanctuary-def has its own test suite.

What does this (Sanctuary) test verify?

```javascript
throws(function() { S.add(1, Infinity); },
       errorEq(TypeError,
               'Invalid value\n' +
               '\n' +
               'add :: FiniteNumber -> FiniteNumber -> FiniteNumber\n' +
               '                       ^^^^^^^^^^^^\n' +
               '                            1\n' +
               '\n' +
               '1)  Infinity :: Number, ValidNumber\n' +
               '\n' +
               'The value at position 1 is not a member of ‘FiniteNumber’.\n'));
```

It verifies two things:

  - the type of `S.add` is `FiniteNumber -> FiniteNumber -> FiniteNumber`; and
  - `Infinity` is not a member of `FiniteNumber`.

`FiniteNumber` is defined and tested in sanctuary-def, so it's not necessary to test it in the Sanctuary test suite as well.

It *is* useful to verify the type of `S.add`, but throwing and catching an exception is an awfully roundabout way of doing so. If `S.add.toString()` were to evaluate to `'add :: FiniteNumber -> FiniteNumber -> FiniteNumber'`, we could replace the test case above with the following:

```javascript
eq(S.add.toString(), 'add :: FiniteNumber -> FiniteNumber -> FiniteNumber');
```

This would also be useful when debugging and when working in the REPL. By defining `inspect` as well as `toString`, one could trivially access a function's type signature:

```
> S.map
map :: Functor f => (a -> b) -> f a -> f b
```

Pretty neat!

### DIFF
--- a/index.js
+++ b/index.js
@@ -1209,6 +1209,17 @@
     }, '').join('\n    ');
   };
 
+  //  typeSignature :: ... -> String
+  var typeSignature = function(
+    name,           // :: String
+    constraints,    // :: StrMap (Array TypeClass)
+    expTypes        // :: Array Type
+  ) {
+    return name + ' :: ' +
+             constraintsRepr(constraints, id, K(K(id))) +
+             expTypes.map(showType).join(' -> ');
+  };
+
   //  _underline :: ... -> String
   var _underline = function recur(
     t,              // :: Type
@@ -1247,9 +1258,7 @@
       };
     }, {carets: [], numbers: [], counter: 0});
 
-    return name + ' :: ' +
-             constraintsRepr(constraints, id, K(K(id))) +
-             expTypes.map(showType).join(' -> ') + '\n' +
+    return typeSignature(name, constraints, expTypes) + '\n' +
            _(name + ' :: ') +
              constraintsRepr(constraints, _, underlineConstraint) +
              st.carets.join(_(' -> ')) + '\n' +
@@ -1468,7 +1477,7 @@
       _indexes,     // :: Array Integer
       impl          // :: Function
     ) {
-      return arity(_indexes.length, function() {
+      var curried = arity(_indexes.length, function() {
         if (checkTypes) {
           var delta = _indexes.length - arguments.length;
           if (delta < 0) {
@@ -1535,6 +1544,9 @@
                        impl);
         }
       });
+      curried.inspect =
+      curried.toString = always(typeSignature(name, constraints, expTypes));
+      return curried;
     };
 
     return function def(name, constraints, expTypes, impl) {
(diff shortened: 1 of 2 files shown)


safareli:
What if `toString` delegates to original function?
```js
add.toString() === 'function(x, y) { return x + y; }'
```

And add method `signature`:
```js
add.signature() === 'add :: Number -> Number -> Number'
```

`toString` could also do both as well 
```js
add.toString() === '/* add :: Number -> Number -> Number */ function(x, y) { return x + y; }'
```

davidchambers:
> What if `toString` delegates to original function?
>
> ```javascript
> add.toString() === 'function(x, y) { return x + y; }'
> ```

Wouldn't this be a bit misleading? I would not expect `add(2)(2)` to work given this string representation.

safareli:
hmm Yes, and it would be even misleading-er as we would not expect `add(_)(2)(_)(2)`  to work that string.
something like this?:
```js
'def(/* add :: Number -> Number -> Number */ function(x, y) { return x + y; })'
// or 
'def("add :: Number -> Number -> Number", function(x, y) { return x + y; })'
// or 
'def(/* Number -> Number -> Number */ function(x, y) { return x + y; })'
// or 
'def("Number -> Number -> Number", function(x, y) { return x + y; })'

```

davidchambers:
Ideally `toString` would guarantee this property:

    forall x. eval(toString(x)) = x

We could start with this:

```javascript
> S.add.toString()
'def("add", {}, [$.FiniteNumber, $.FiniteNumber, $.FiniteNumber], function(a, b) { return a + b; })'
```

We can't `eval` the result without knowing what `def` is. We'd need to replace `def` with its definition:

```javascript
> S.add.toString()
'$.create({checkTypes: true, env: […]})("add", {}, [$.FiniteNumber, $.FiniteNumber, $.FiniteNumber], function(a, b) { return a + b; })'
```

We'd then need to replace `…` with the names of all the types in the environment. The result would be so noisy that I'd prefer a type-level representation of the function nine times out of ten.

Whatever we do will be an improvement on the current behaviour:

```javascript
> S.add.toString()
'function ($1, $2) {\n          return f.apply(this, arguments);\n        }'
```

davidchambers:
I'm going to merge this pull request in its current state. The `inspect` and `toString` methods are undocumented; we're certainly free to change them if we wish to do so. I predict we will :heart: quick access to type signatures in the REPL, so we may decide that `inspect` should remain as is even if we decide to change `toString`.

PR #89, opened 2016-10-22
enforce type-variable constraints for record types

### LINKED ISSUE #35: record types ignore type-variable constraints
Given:

```javascript
//    a :: Type

//    Foo :: { x :: a, y :: a }
const Foo = $.RecordType({x: a, y: a});

//    def :: (String, StrMap [Type], [Type], Function) -> Function
const def = $.create(true, $.env.concat([Foo]));

//    id = Foo -> Foo
const id = def('id', {}, [Foo, Foo], x => x);
```

This should be a type error:

```javascript
id({x: 'abc', y: 123});
```

As should this:

```javascript
id({x: ['foo', false], y: ['foo', false]});
```

davidchambers:
Closes #35

### DIFF
--- a/index.js
+++ b/index.js
@@ -1012,6 +1012,16 @@
             });
           });
 
+        case RECORD:
+          return expType.keys.reduce(function(e, k) {
+            return e.chain(function(r) {
+              return recur(r.typeVarMap,
+                           expType.types[k].type,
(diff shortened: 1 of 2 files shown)


davidchambers:
Thanks for having a look at this patch, @safareli and @Avaq. :)

PR #90, opened 2016-10-22
only determine actual types when necessary

### LINKED ISSUE #16: Unnecessary Invocations Of Type#test Functions
I noticed something strange last night, it seems that type checking is performed for every type in the environment for a function regardless of whether that type is part of a function's type signature or not.

E.g.,
```js
const $test = $.NullaryType(
  'TEST',
  x => typeof x === 'number'
);

const $notUsed = $.NullaryType(
  'NOTUSED',
  x => {
    console.log('Why am I being used?');
    return typeof x === 'string'
  }
);

const env = $.env.concat([$test, $notUsed]);
const def = $.create(env);

const f = def('f', {}, [$test, $test], x => x + 1);
f(1);
// Why am I being used?
// Why am I being used?
```

If I put a `console.log` in `NullaryType`'s `test` function I see that there is an invocation for each one of those types too and I'm guessing for any other. This seems quite wasteful to me and possibly a source of confusion. In fact, the only reason I found this issue was because I was getting this cryptic error message for a custom type I'd defined which I realised, after much investigation, had a `test` function with a call that caused one of the built in types to throw even though I wasn't using that type.

### LINKED ISSUE #76: pay cost of type variables only for polymorphic functions
For `a -> a` it's necessary to determine all the types of which the argument is a member in order to determine whether the return value is valid. For `String -> Integer`, though, all we need to do is determine whether the argument is a member of `String`.

Given a million-element array we'll currently determine the types of each array element, then find the intersection of these million arrays! We could use a more efficient approach in the `a -> a` case, but the most efficient approach in the `String -> Integer` case is certainly `$.String._test(x)`.

davidchambers:
Closes #16; closes #76

Commit message:

> In order to reach the default case `expType` must be an enumerable type, a function type, or a nullary type. It therefore contains no immediately resolvable type variables, so we needn't determine all the types of which all the given values are members.

I'm slightly embarrassed by how straightforward this fix turned out to be! With this project I often need to spend several hours loading everything into (my) memory before becoming productive. :bulb:

### DIFF
--- a/index.js
+++ b/index.js
@@ -1023,8 +1023,7 @@
           }, Right({typeVarMap: typeVarMap, types: [expType]}));
 
         default:
-          return Right({typeVarMap: typeVarMap,
-                        types: determineActualTypesStrict(env, env, values)});
+          return Right({typeVarMap: typeVarMap, types: [expType]});
       }
     };
   };
--- a/test/index.js
+++ b/test/index.js
@@ -2557,6 +2557,29 @@ describe('def', function() {
            'The value at position 1 is not a member of ‘FiniteNumber’.\n');
   });
 
+  it('only determines actual types when necessary', function() {
+    //  count :: Integer
+    var count = 0;
+
+    //  Void :: Type
+    var Void = $.NullaryType('my-package/Void', function(x) { count += 1; return false; });
+
+    var env = [$.Array, Maybe, $.Number, Void];
+    var def = $.create({checkTypes: true, env: env});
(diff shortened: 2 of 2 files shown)


Avaq:
Oh wow. Straight forward indeed! Were you able to observe a difference in performance? :no_mouth:

davidchambers:
> Were you able to observe a difference in performance?

No, surprisingly!

PR #104, opened 2016-11-27
use function declarations

davidchambers:
Evaluation order is a significant source of complexity in computer programs<sup>†</sup>.

[Function declaration hoisting][1] reduces the significance of ordering. It is thus—contrary to popular opinion—one of JavaScript's *good* parts. In fact, it would be terrific if *all* assignments were hoisted. `const x = 2 * y; const y = 21;` should be a valid program. :)

I was motivated to make this change while working on a feature. I found myself frequently changing functions from expressions to declarations. It should be possible to rearrange the function definitions, of course, but this would add a lot of noise to the diff and restrict our ability to order definitions with the [Transcribe][2] output in mind.

---

† I read a paper which identified ordering as a source of program complexity. I can't remember its title or the name of its author, and I haven't been able to locate it via Google. If you happen to know the paper, please post a comment. :)

[1]: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Statements/function#Function_declaration_hoisting
[2]: https://github.com/plaid/transcribe

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ lint:
 	  --global module \
 	  --global require \
 	  --global self \
-	  --rule 'func-style: [off]' \
+	  --rule 'func-style: [error, declaration]' \
 	  --rule 'max-len: [error, {code: 79, ignorePattern: "^ *//(#|[.] //) ", ignoreUrls: true}]' \
 	  -- index.js
 	$(ESLINT) \
--- a/index.js
+++ b/index.js
@@ -191,7 +191,7 @@
   var toString          = Object.prototype.toString;
 
   //  Left :: a -> Either a b
-  var Left = function Left(x) {
+  function Left(x) {
     return {
       '@@type': 'sanctuary-def/Either',
       isLeft: true,
@@ -200,10 +200,10 @@
       'fantasy-land/map': function(f) { return Left(x); },
       value: x
     };
-  };
+  }
 
   //  Right :: b -> Either a b
-  var Right = function Right(x) {
+  function Right(x) {
     return {
       '@@type': 'sanctuary-def/Either',
       isLeft: false,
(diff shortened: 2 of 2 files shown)


svozza:
I prefer function declarations too although I can't agree with this:

>In fact, it would be terrific if all assignments were hoisted. const x = 2 * y; const y = 21; should be a valid program. :)

Having been caught in production with bugs caused by `var` hoisting I am glad that JavaScript now has properly block scoped variable declarations like every other programming language.

davidchambers:
> > In fact, it would be terrific if all assignments were hoisted. `const x = 2 * y; const y = 21;` should be a valid program. :)
>
> Having been caught in production with bugs caused by `var` hoisting I am glad that JavaScript now has properly block scoped variable declarations like every other programming language.

What I'm suggesting is very different from JavaScript's `var` hoisting, which just means that…

```javascript
function main() {
  var x = 2 * y;
  var y = 21;
  console.log(x);
}
```

becomes…

```javascript
function main() {
  var x, y;
  x = 2 * y;
  y = 21;
  console.log(x);
}
```

which I agree is harmful rather than useful.

I :heart: languages which don't care about the order in which I define things. Consider this trivial Haskell program, for example:

```haskell
x = 2 * y
y = 21

main = putStrLn $ show x
```

```console
$ ghc example.hs && ./example
[1 of 1] Compiling Main             ( example.hs, example.o )
Linking example ...
42
```

safareli:
Also function declarations need no semicolons 😄 


>x = 2 * y
>y = 21

verbose and inefficient version 😃 

```js
console.log(x())
function x(){ return 2 * y()}
function y(){ return 21}
```

svozza:
> What I'm suggesting is very different from JavaScript's var hoisting

Aha, I was wondering! It's funny now that you mention it, I never even consider definition order when I write Haskell.

davidchambers:
I'd been thinking along similar lines, @safareli. Thunking everything (using function declarations) provides a reasonable approximation of lazy evaluation. :)

PR #105, opened 2016-11-30
add ‘url’ parameter to type constructors to improve error messages

davidchambers:
```javascript
//    inc :: FiniteNumber -> FiniteNumber
const inc = def('inc', {}, [$.FiniteNumber, $.FiniteNumber], x => x + 1);

inc(Infinity);
// ! TypeError: Invalid value
//
//   inc :: FiniteNumber -> FiniteNumber
//          ^^^^^^^^^^^^
//               1
//
//   1)  Infinity :: Number
//
//   The value at position 1 is not a member of ‘FiniteNumber’.
//
//   See https://github.com/sanctuary-js/sanctuary-def/tree/v0.7.0#FiniteNumber for information about the sanctuary-def/FiniteNumber type.
```

The link to the type's documentation strikes me as useful. What do you think? It makes defining types a bit less convenient, but provides a benefit to those who use functions defined via sanctuary-def.

Using version-specific URLs is a good idea as a type may be deleted or its meaning may change.

### DIFF
--- a/index.js
+++ b/index.js
@@ -310,6 +310,7 @@
   function createType(
     typeName,   // :: String
     name,       // :: String
+    url,        // :: String
     format,     // :: (String -> String, String -> String -> String) -> String
     test,       // :: Any -> Boolean
     keys,       // :: Array String
@@ -341,6 +342,7 @@
       toString: always(format(id, K(id))),
       type: typeName,
       types: types,
+      url: url,
       validate: validate
     };
   }
@@ -357,7 +359,7 @@
 
   //  Inconsistent :: Type
   var Inconsistent =
-  createType(INCONSISTENT, '', always2('???'), K(false), [], {});
+  createType(INCONSISTENT, '', '', always2('???'), K(false), [], {});
 
   //  $$type :: a -> String
   function $$type(x) {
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -2135,15 +2161,18 @@
...
       'Invalid value\n\n' +
       underline(typeInfo,
                 K(K(_)),
                 formatType6(Z.concat([index], propPath))) +
       '\n' +
       showValuesAndTypes(env, [value], 1) + '\n\n' +
-      'The value at position 1 is not a member of ' +
-      showTypeQuoted(resolvePropPath(typeInfo.types[index], propPath)) + '.\n'
+      'The value at position 1 is not a member of ' + s + '.\n' +
+      (t.url &&
```
davidchambers: This means one can simply use `''` as the `url` of an undocumented type.

index.js
```diff
@@ -380,17 +382,28 @@
     };
   }
 
+  function typeX(name, test, _1, _2) {
+    var version = '0.7.0';  // updated programmatically
```
safareli: can we get it from package.json

index.js
```diff
@@ -380,17 +382,28 @@
...
+    var url = 'https://github.com/sanctuary-js/sanctuary-def/tree/v' +
+              version + '#' + stripNamespace(name);
+    switch (arguments.length) {
```
safareli: What about something like this? (this names are not perfect)
```js
$.FiniteNumber = withURL('sanctuary-def/FiniteNumber', (name, url) => NullaryType(
  name, url,
  function(x) { return $.ValidNumber._test(x) && isFinite(x); }
));
// or 
$.FiniteNumber = typeWithURL(
  'sanctuary-def/FiniteNumber',
  NullaryType,
  function(x) { return $.ValidNumber._test(x) && isFinite(x); }
);
```
Whis way we would not need this switch statement. And it will eliminate need to update it if we add some other `*Type` with more arguments. Also it will still be clear which type a value is (user will scan the definition and see `NullaryType` or whatever immediately instead of counting number of arguments.


```js
$.Pair = typeWithURL(
  'sanctuary-def/Pair',
  BinaryType,
  function(x) { return $$typeEq('Array')(x) && x.length === 2; },
  function(pair) { return [pair[0]]; },
  function(pair) { return [pair[1]]; }
)
// vs
$.Pair = typeX(
  'sanctuary-def/Pair',
  function(x) { return $$typeEq('Array')(x) && x.length === 2; },
  function(pair) { return [pair[0]]; },
  function(pair) { return [pair[1]]; }
)
```

index.js
```diff
@@ -1379,15 +1403,15 @@
   //. //    Rank :: Type
   //. const Rank = $.NullaryType(
   //.   'my-package/Rank',
-  //.   x => typeof x === 'string' && /^([A23456789JQK]|10)$/.test(x),
-  //.   'A'
```
safareli: don't know what this `'A'` and `\u2660` are, but is it intentional?

index.js
davidchambers: Not in the browser, unfortunately.

index.js
davidchambers: Great suggestion, Irakli! I was not at all happy with `typeX`. I've updated the pull request. :)

safareli:
👍

PR #106, opened 2016-12-02
use ‘def’ to define exported functions, for currying and type checking

### LINKED ISSUE #102: Error Message String Contains [object Object]
I'm currently working on the PR generated by https://github.com/sanctuary-js/sanctuary/issues/291. The new implementation of `pluck` looks like this:

```js
def('pluck',
      {a: [Accessible]},
      [$.String, $.Array({a: [Accessible]}), $.Array(a)],
      pluck);
```

Unfortunately it seems that `sanctuary-def` can't handle `$.Array({a: [Accessible]})`:

```
TypeError: Invalid value

pluck :: Accessible a => String -> Array [object Object] -> Array a
                         ^^^^^^
                           1
```

davidchambers:
Closes #102

With this change you'd have encountered a helpful error message, @svozza:

```javascript
$.Array({a: [Accessible]});
// ! TypeError: Invalid value
//
//   Array :: Type -> Type
//            ^^^^
//             1
//
//   1)  {"a": [[object Object]]} :: Object, StrMap (Array ???)
//
//   The value at position 1 is not a member of ‘Type’.
```

[`TypeClass`][1] values having `'[object Object]'` as their string representation is not ideal, of course, but at least we're now rejecting invalid arguments to the `$.Array` type constructor. :)

[1]: https://github.com/sanctuary-js/sanctuary-type-classes#TypeClass

### DIFF
--- a/index.js
+++ b/index.js
@@ -159,8 +159,6 @@
 
   'use strict';
 
-  var $ = {};
-
   //# __ :: Placeholder
   //.
   //. The special placeholder value.
@@ -180,12 +178,11 @@
   //. exclaim('ahoy');
   //. // => 'ahoy!'
   //. ```
-  $.__ = {'@@functional/placeholder': true};
+  var __ = {'@@functional/placeholder': true};
 
   var MAX_SAFE_INTEGER = Math.pow(2, 53) - 1;
   var MIN_SAFE_INTEGER = -MAX_SAFE_INTEGER;
 
-  var forEach           = Array.prototype.forEach;
   var slice             = Array.prototype.slice;
   var hasOwnProperty    = Object.prototype.hasOwnProperty;
   var toString          = Object.prototype.toString;
@@ -411,7 +408,7 @@
   function applyParameterizedTypes(types) {
     return Z.map(function(x) {
       return typeof x === 'function' ?
-        x.apply(null, Z.map(K($.Unknown), range(0, x.length))) :
+        x.apply(null, Z.map(K(Unknown), range(0, x.length))) :
         x;
     }, types);
   }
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1582,22 +1630,6 @@
...
-        'The argument to ‘RecordType’ must be an object ' +
-          'mapping field name to type.\n\n' +
-        'The following mappings are invalid:\n\n' +
-        Z.reduce(function(s, k) {
-          var v = fields[k];
-          return s + '  - ' + Z.toString(k) + ': ' + Z.toString(v) + '\n';
-        }, '', invalidFieldNames)
-      ));
-    }
-
```
davidchambers: Replacing this with `def('RecordType', {}, [StrMap(Type), Type], RecordType)` feels very good. :)

davidchambers:
I rewrote the one part of the patch that bothered me. I'd appreciate a review (even a cursory one).

test/index.js
```diff
@@ -1193,14 +1237,15 @@ describe('def', function() {
...
-           'Invalid values\n' +
+           'Invalid value\n' +
            '\n' +
-           'The argument to ‘RecordType’ must be an object mapping field name to type.\n' +
+           'RecordType :: StrMap Type -> Type\n' +
+           '                     ^^^^\n' +
+           '                      1\n' +
            '\n' +
-           'The following mappings are invalid:\n' +
+           '1)  /XXX/ :: RegExp\n' +
```
safareli: is is sort of degradation? before both `x` and `y` was shown in error but now only `x`

index.js
```diff
@@ -1173,7 +1175,7 @@
     }
   }
 
-  //# test :: (Array Type, Type, a) -> Boolean
+  //# test :: Array Type -> Type -> a -> Boolean
```
safareli: why have changed the signature to curried form? looks like the `test` function is not curried

index.js
```diff
@@ -1944,11 +1979,16 @@
     });
   }
 
-  //  invalidArgumentsLength :: (String, Integer, Integer) -> Error
-  function invalidArgumentsLength(name, expectedLength, actualLength) {
-    return new TypeError(q(name) +
-                         ' requires ' + numArgs(expectedLength) + ';' +
-                         ' received ' + numArgs(actualLength));
+  //  tooManyArguments :: (TypeInfo, Integer) -> Error
```
safareli: can we use `invalidArgumentsLength` instead of adding this `tooManyArguments` function?

safareli:
have not looked into `curry` and `create_` changes

test/index.js
davidchambers: Currently `RecordType` has a 15-line block of custom type checking, which allows it to provide customized error messages. The new error messages are slightly less informative, but they are more consistent with the error messages produced in other contexts. On balance I think replacing the custom type checking with

```javascript
def('RecordType', {}, [StrMap(Type), Type], RecordType)
```

is worthwhile. What do you think?

index.js
davidchambers: The internal `test` function is not curried, but the exported `test` function is:

```javascript
return {
  // ...
  test: def('test', {}, [Array_(Type), Type, Any, Boolean_], test),
  // ...
};
```

Having this function curried is very convenient. I can imagine partially applying it to fix the environment:

```javascript
//    env :: Array Type
const env = ...;

//    test :: Type -> a -> Boolean
const test = $.test(env);
```

test/index.js
safareli: ah i see, agree on this. Maybe in future we could update the `def` to have same results (listing all issues instead of just first one).

index.js
safareli: did not noticed that

index.js
davidchambers: This is a bit confusing. Currently we have two functions:

  - `invalidArgumentsLength`, used when a `def`'d function itself is misapplied; and
  - `invalidArgumentsLength_`, used when a function *argument* is misapplied.

The difference is best explained by example:

```javascript
const $ = require('.');

const b = $.TypeVariable('b');

const def = $.create({checkTypes: true, env: $.env});

//    map :: (a -> b) -> Array a -> Array b
const map =
def('map',
    {},
    [$.Function([a, b]), $.Array(a), $.Array(b)],
    (f, xs) => xs.map(f));

map(Math.sqrt, [], [], [], []);
// ! TypeError: ‘map’ requires two arguments; received five arguments

map(Math.sqrt, [1, 4, 9]);
// ! TypeError: ‘map’ applied ‘a -> b’ to the wrong number of arguments
//
//   map :: (a -> b) -> Array a -> Array b
//           ^
//           1
//
//   Expected one argument but received three arguments:
//
//     - 1
//     - 0
//     - [1, 4, 9]
```

`invalidArgumentsLength` is currently referenced in two places.

In `curry`:

```javascript
var delta = _indexes.length - arguments.length;
if (delta < 0) {
  throw invalidArgumentsLength(typeInfo.name, n, n - delta);
}
```

In `def`:

```javascript
if (arguments.length !== def.length) {
  throw invalidArgumentsLength('def', def.length, arguments.length);
}
```

Now that `$.def` is defined via `def`, the latter is no longer necessary. In `curry` we only ever invoke `invalidArgumentsLength` when we receive too *many* arguments, so we can make the error message more specific (and change the function's name from `invalidArgumentsLength` to `tooManyArguments`).

If we're to rename `invalidArgumentsLength`, the `_` in `invalidArgumentsLength_` is no longer necessary. This pull request makes the following changes:

  - `invalidArgumentsLength` → `tooManyArguments`; and
  - `invalidArgumentsLength_` → `invalidArgumentsLength`.

Have I done a reasonable job of making this confusing situation clearer?

safareli: now it's all clear!

davidchambers: Great! Thank you, Irakli, for prompting me to explain these changes. Recording such things is important, but I will often neglect to do so unless someone asks a question.

PR #110, opened 2016-12-19
test: add assertions for ‘name’ and ‘url’ properties

davidchambers:
Some of these assertions should have been added in #105.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -1356,7 +1356,90 @@ describe('def', function() {
            'Since there is no type of which all the above values are members, the type-variable constraint has been violated.\n');
   });
 
-  it('supports the "ValidDate" type', function() {
+  it('provides the "Any" type', function() {
+    eq($.Any.name, 'sanctuary-def/Any');
+    eq($.Any.url, 'https://github.com/sanctuary-js/sanctuary-def/tree/v' + version + '#Any');
+  });
+
+  it('provides the "AnyFunction" type', function() {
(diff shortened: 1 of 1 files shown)


safareli:
btw what about something like this?
```js
function run(dec) {
  Object.keys(dec).forEach(function(k){
    it('provides the "'+k+'" type', function() {
      eq(dec[k].name, k);
      eq(dec[k].url, 'https://github.com/sanctuary-js/sanctuary-def/tree/v' + version + '#' + k);
    });
  })
}

run({
  Array: $.Array(a),
  Boolean: $.Boolean,
  ...
})
```

davidchambers:
That's certainly an option. I don't mind repetition in a test suite, though. In this case it keeps things simple as we don't need to have `run` handle both prefixed (e.g. `'sanctuary-def/FiniteNumber'`) and unprefixed (e.g. `'Number'`) type identifiers.

The main reason I added these assertions is that in several cases I don't like the current values. By documenting the current behaviour we will get informative diffs if and when we decide to change values. Here's part of the diff from #111, for example:

```diff
-    eq($.RegexFlags.name, '');
-    eq($.RegexFlags.url, '');
+    eq($.RegexFlags.name, 'sanctuary-def/RegexFlags');
+    eq($.RegexFlags.url, 'https://github.com/sanctuary-js/sanctuary-def/tree/v' + version + '#RegexFlags');
```

PR #114, opened 2016-12-28
Changed 'add' to the less generic 'subtr'

dmitriz:
I suggest the subtraction 'subtr' as a more natural example for enforcing two arguments, as opposed to the 'add' function, for which it actually makes more sense to allow arbitrary number of arguments.


README.md
```diff
@@ -87,31 +87,31 @@ JavaScript's implicit type coercion often obfuscates the source of type
 errors. Consider the following function:
 
 ```javascript
-//    _add :: (Number, Number) -> Number
-const _add = (x, y) => x + y;
+//    _subtr :: (Number, Number) -> Number
+const _subtr = (x, y) => x + y;
```
safareli: `-`

README.md
```diff
@@ -87,31 +87,31 @@ JavaScript's implicit type coercion often obfuscates the source of type
...
 ```
 
-The type signature indicates that `_add` takes two arguments of type
+The type signature indicates that `_subtr` takes two arguments of type
 `Number`, but this is not enforced. This allows type errors to be silently
 ignored:
 
 ```javascript
-_add('2', '2');
+_subtr('2', '2');
```
safareli:
```js
'2' - '2' // 0
```

dmitriz:
Sorry missed some parts, hopefully it is correct now.

safareli:
Also the `README.md` is generated from comments in `index.js` so i think this changes should be moved there.

dmitriz:
Hm... just noticed that the same Readme is inside `index.js`, so that was prob the wrong file to change if it is auto-generated?

safareli:
You should only change comments in index.js, README is generated automatically on release of each version

dmitriz:
Just made local changes to the comments in `index.js` but running `npm test` does not seem to update `README.md`. 

Am I missing something?

dmitriz:
Here is the changed file:
https://github.com/dmitriz/sanctuary-def/blob/subtr-replaces-add/index.js

safareli:
Tests is not updating `README`,  it's updated when new version is released, you shouldn't change it.



`//. const inc = subtrFrom 10(10);` extra space after `subtrFrom`.
You should move the index.js changes to this `patch-1` branch

dmitriz:
@safareli Thanks, all is hopefully fixed. BTW, should `README.md` not go into `.gitignore` it is auto-generated?

yarn.lock
```diff
@@ -0,0 +1,1107 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
safareli: this file should be removed from this PR

safareli:
If file is autogenerated, it doesn't automatically mean, that it should be ignored, if we ignore README then the repo will have no description on github and on npm.

yarn.lock
dmitriz: At Yarn, they recommend to check it:
https://yarnpkg.com/en/docs/version-control#toc-required-files

Is there any reason not to follow their recommendation?

dmitriz:
@safareli 
>If file is autogenerated, it doesn't automatically mean, that it should be ignored, if we ignore README then the repo will have no description on github and on npm.

Good point 😄

yarn.lock
Shard: Yarn is currently being considered in https://github.com/sanctuary-js/sanctuary/issues/265

davidchambers:
Here's the crux of the change:

```diff
-//. add(2, 2, 2);
-//. // ! TypeError: ‘add’ requires two arguments; received three arguments
+//. subtr(6, 2, 1);
+//. // ! TypeError: ‘subtr’ requires two arguments; received three arguments
```

> the 'add' function […] actually makes more sense to allow arbitrary number of arguments

I agree wholeheartedly (although [`clojure.core/-`][1] is variadic). The reason I chose `add`, though, is that it's a function that could be variadic but in this case is not. One would expect `add(2, 2, 2)` to evaluate to `6` rather than `4`. I believe that applying a function to too many arguments should crash the program; `add(2, 2, 2)` is a compelling example of the danger of silently ignoring superfluous arguments. Although I'm a firm believer in the merits of fixed-arity functions, that's a more nuanced matter relating to API design rather than correctness.

I appreciate the contribution, @dmitriz, but I believe this change weakens rather than strengthens the message I intended to convey.


[1]: https://clojuredocs.org/clojure.core/-

dmitriz:
@davidchambers Thanks, I can see your point, then perhaps it should be called `binaryAdd` instead of `add`?  That would reveal the intent of limiting the number of arguments and, possibly, underline your point even stronger?

As liberal polymorphist 😄 I would reserve the more general `add` for arbitrary number of arguments, because that is how it is used and taught in maths. Each has its use cases of course.

Good point about the variadic `-` in Closure, interesting, actually makes sense: 

    10 - 3 - 2 // => 5

dmitriz:
@davidchambers I see, thanks for explaining it. My perspective is slightly different though 😄 :

If you tell me about a problem that I can solve myself and am happy with it, I won't be convinced. So if I see the value to use Hungarian notation in my code and feel it works fine for me, I will be less tempted to look for other ways.

In our example, the additional value I am looking for from this library is to make my code safer without manually polluting my code with type checks and guards (which I really hate). At the same time, I am striving for the best possible readability. The two goals are different. With `binaryAdd` I improve the readability, but I still want a safety check, in case it is used wrongly. The suggestive naming `binaryAdd` alone does not help me at all to sleep better at night. 😄 
So even though, it is a correct name, I still have the problem that I cannot fix by better naming.

However, when I read "`add(2, 2, 2)` is wrong", my immediate reaction is -- why? That is, what normally the function does after all. So now, instead of focusing on the value of the library to guard my safety, I am distracted by the thought, why would I need it in first place. Why trying to change the perfectly fine working function? As a newcomer, that may put me in the wrong thinking direction, instead of focusing and what the library brings to the table.

davidchambers:
Ah, so you're saying someone might look at `add(2, 2, 2)` and think the function's author missed an opportunity to make the function more flexible by having it operate on exactly two inputs?

Perhaps what is missing is an explanation of the connection between variadic functions and currying: we cannot have the former if we want the latter.

A question to ponder is to what extent the documentation should sell the library. It should provide motivating examples, of course, but reasons to use the library at all may be better conveyed in blog posts, screencasts, and meetup/conference talks. @JAForbes, for example, is a good advocate of functional programming in the JavaScript community.

dmitriz:
> Ah, so you're saying someone might look at add(2, 2, 2) and think the function's author missed an opportunity to make the function more flexible by having it operate on exactly two inputs?

Exactly :)

> Perhaps what is missing is an explanation of the connection between variadic functions and currying: we cannot have the former if we want the latter.

Indeed. But people will want to use both as each has its advantages.


> A question to ponder is to what extent the documentation should sell the library. It should provide motivating examples, of course, but reasons to use the library at all may be better conveyed in blog posts, screencasts, and meetup/conference talks. @JAForbes, for example, is a good advocate of functional programming in the JavaScript community.

I was unaware of @JAForbes work, thanks for mentioning it! But I suppose you cannot avoid people coming with question like "why should I use it?", "what does it bring to the table?", so it is always good to give some motivation (i.e. to sell 😄 ) right at the beginning of the Readme, as that is what people read first. As reader, I always appreciate when effort is made to address those questions.

davidchambers:
This pull request prompted quite a few comments, @dmitriz. :)

I was already leaning towards closing this—from time to time a (benevolent) dictator's gotta dictate—but seeing this convinced me it's the right move:

```javascript
//    _subtr :: (Number, Number) -> Number
const _subtr = (x, y) => x + (-y);
```

I've never seen a subtraction function defined this way. ;)

This thread has made me appreciate just how *well* `_add` and `add` perform their roles: highlighting the dangers of type coercion (`_add('2', '2') = '22'`) and of ignoring superfluous arguments (`add(2, 2, 2) = 4`), and providing a nice clear example of partial application (`inc = add(1)`).

> But people will want to use both as each has its advantages.

What are the advantages of variadic functions? I suppose `Math.max` can act as both a binary function and a function for finding the largest value in an array, though the latter is cumbersome. Far better, in my opinion, to define two functions (as Haskell does):

```purescript
max :: Ord => a -> a -> a
maximum :: Ord a => [a] -> a
```

I will close this pull request, but I don't intend to suppress further comments.

dmitriz:
@davidchambers I mostly thank for your time and patience to explain the things, and can now understand better the reasons behind your decisions.

> I've never seen a subtraction function defined this way. ;)

Yeah, it was a bit creative ;)
On a deeper level, it is actually a way of reducing complexity with lower coupling and thinner interfaces. The plus `+` is already provided, and unary `-` is simpler than its binary "sister", no real need to import another binary function. ;) 

>What are the advantages of variadic functions? I suppose Math.max can act as both a binary function and a function for finding the largest value in an array, though the latter is cumbersome. Far better, in my opinion, to define two functions (as Haskell does):
```js
max :: Ord => a -> a -> a
maximum :: Ord a => [a] -> a
```

From my perspective as mathematician, Haskell's special treatment of pairs vs other n-tuples is actually confusing and seemingly inconsistent:
http://stackoverflow.com/questions/41327198/applying-lifted-functions-to-tuples-of-arbitrary-length-in-haskell

davidchambers:
> I mostly thank for your time and patience to explain the things, and can now understand better the reasons behind your decisions.

That's very good to hear. :)

> Haskell's special treatment of pairs vs other n-tuples is actually confusing and seemingly inconsistent

I don't see the connection, I'm afraid. Neither [`max`][1] nor [`maximum`][2] uses the `(,)` type constructor.


[1]: https://hackage.haskell.org/package/base-4.9.0.0/docs/Prelude.html#v:max
[2]: https://hackage.haskell.org/package/base-4.9.0.0/docs/Prelude.html#v:maximum

dmitriz:
> Haskell's special treatment of pairs vs other n-tuples is actually confusing and seemingly inconsistent

> I don't see the connection, I'm afraid. Neither max nor maximum uses the (,) type constructor.

Not directly, but `f(x,y)` can be mathematically identified with `f` applied to the pair `(x, y)`. I know this is not implemented in Haskell, but that is, in my view, a kind of overloading that is rather natural, convenient and unambiguous. This is btw closely related to how JavaScript handles its function arguments -- with the `arguments` object, which behaves very closely to a tuple as far as I can see. One can argue that array can be used, but the Haskell's restriction on array argument types makes them unsuitable, whereas tuples seems exactly right.

Concerning the variadic `max` and `add` -- that is actually a very convenient overloading and is precisely how it is used in maths with no confusion. In fact, any associative binary operation is naturally extendible to a variadic function, for which you would never need a new name. Without it, you would have to compose nested `add`s, which would easily become unreadable.

BTW, Ramda's `pipe` and `compose` are very convenient variadic functions ;)

Also I personally by far prefer a single polymorphic  function `max` over using two different names, which btw do not give me any clue about which one is to use. That sounds like a path to nightmare, constantly mixing them up and getting errors. You have no problem with the polymorphic `map` method and function, so I fail to see any problem with `max` behaving the same way (i.e. working on both tuples and arrays). 

My opinion may not count though, as I am not a Haskell user and may not know other parts of the story...

davidchambers:
Your comment above helped me to better understand your position. Thank you.

> BTW, Ramda's `pipe` and `compose` are very convenient variadic functions ;)

Indeed. It's worth noting, though, that neither [`S.pipe`][1] nor [`S.compose`][2] is variadic. ;)

> I fail to see any problem with `max` behaving the same way (i.e. working on both tuples and arrays)

We get to chose between this and currying. I choose currying. I want to be able to define the _at least zero_ function via partial application: `max 0`.

As for the names, I agree that having two similar names for similar functions is unfortunate. JavaScript's `String#substr` and `String#substring` have long bothered me!


[1]: https://sanctuary.js.org/#pipe
[2]: https://sanctuary.js.org/#compose

dmitriz:
> Your comment above helped me to better understand your position. Thank you.

You are welcome! 😄 

>>BTW, Ramda's pipe and compose are very convenient variadic functions ;)
>Indeed. It's worth noting, though, that neither S.pipe nor S.compose is variadic. ;)

LOL. Of course one could argue, it is closely related (isomorphic) as you basically replace the arbitrary number of args by arrays of arbitrary length. 

BTW, a notable syntactic advantage of the array approach is the trailing comma tolerance, whose absence is something that annoys me with variadic functions ;)
However, that feels like a superficial issue related to the JS current implementation, 
like earlier it used not to allow those commas in the arrays as well,
and perhaps there is already (or should be) a transpiler removing those commas.

This has no array - is it correct?
https://github.com/sanctuary-js/sanctuary/blob/v0.11.1/index.js#L706
```js
  // compose :: (b -> c) -> (a -> b) -> a -> c
```

>>I fail to see any problem with max behaving the same way (i.e. working on both tuples and arrays)
> We get to chose between this and currying. I choose currying. I want to be able to define the at least zero function via partial application: max 0.

You basically use the binary part. It can be called `max2` (in analogy with `curryN`). Not beautiful but expressive, unambiguosly clear and easy to remember. Then you may want to use `max3`, why limiting your life to counting till 2? ;)

> As for the names, I agree that having two similar names for similar functions is unfortunate. JavaScript's String#substr and String#substring have long bothered me!

Indeed, a terrible naming choice. I always encapsulate such things into something more meaningful to have a sane life ;)

PR #121, opened 2017-02-23
Correct type of $.test

Avaq:
Opened the WIP pull-request for discussion. Comments will follow.

### DIFF
--- a/index.js
+++ b/index.js
@@ -762,6 +762,15 @@
     typeEq('sanctuary-def/Type')
   );
 
+  //  VariadicType :: Type | Type -> Type | Type -> Type -> Type | ...
+  var VariadicType = NullaryType(
+    'sanctuary-def/VariadicType',
+    '',
+    function(x) {
+      return test([], Type, x) || test([], AnyFunction, x);
+    }
+  );
+
   //  TypeClass :: Type
   var TypeClass = NullaryType(
     'sanctuary-type-classes/TypeClass',
@@ -2433,7 +2442,7 @@
     ValidNumber: ValidNumber,
     env: env,
     create: create,
-    test: def('test', {}, [Array_(Type), Type, Any, Boolean_], test),
+    test: def('test', {}, [Array_(VariadicType), Type, Any, Boolean_], test),
     NullaryType: CheckedNullaryType,
     UnaryType: CheckedUnaryType,
     BinaryType: CheckedBinaryType,
--- a/test/index.js
+++ b/test/index.js
@@ -2862,7 +2862,7 @@ describe('test', function() {
   it('is a ternary function', function() {
     eq(typeof $.test, 'function');
     eq($.test.length, 3);
-    eq($.test.toString(), 'test :: Array Type -> Type -> Any -> Boolean');
+    eq($.test.toString(), 'test :: Array VariadicType -> Type -> Any -> Boolean');
   });
 
   it('supports nullary types', function() {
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -408,13 +408,16 @@
     return BinaryType(name, functionUrl(name), test, _1, _2);
   }
 
+  //  applyParameterizedType :: (Type... -> Type) -> Type
+  function applyParameterizedType(type) {
+    return typeof type === 'function' ?
+      type.apply(null, Z.map(K(Unknown), range(0, type.length))) :
+      type
+  }
+
```
Avaq: See the reason for this diff in my other comment

index.js
```diff
@@ -762,6 +765,14 @@
     typeEq('sanctuary-def/Type')
   );
 
+  var VariadicType = NullaryType(
+    'sanctuary-def/VariadicType',
+    '',
+    function(x){
+      return test([], Type, x) || test([], AnyFunction, x);
```
Avaq: I tried different ways of testing:

1. `test([], Type, applyParameterizedType(x))` - This blows up if someone would provide, say `[x => true]` as `env`.
1. `test([], Type, x) || test([], Function([Type, Type]), x) || test([], Function([Type, Type, Type]), x)` - This effectively behaves the same as `AnyFunction`, because `test` does not apply `x` to determine its types.
1. This left me with the method seen above.

The decision, really, is between `1` and `3`, where `3` is overly permissive, and `1` blows up with an ugly error. Maybe to get the best of both worlds, I should combine them as:

```js
test([], Type, x) || test([], AnyFunction, x) && test([], Type, applyParameterizedType(x))
```

index.js
```diff
@@ -2405,7 +2416,7 @@
     Date: Date_,
     Error: Error_,
     FiniteNumber: FiniteNumber,
-    Function: def('Function', {}, [Array_(Type), Type], Function_),
+    Function: def('Function', {}, [Array_(VariadicType), Type], Function_),
```
Avaq: I changed this one, but I'm not sure that I should. I'll ask you first. :)

index.js
```diff
@@ -762,6 +765,14 @@
     typeEq('sanctuary-def/Type')
   );
 
+  var VariadicType = NullaryType(
+    'sanctuary-def/VariadicType',
+    '',
+    function(x) {
+      return test([], Type, x) || test([], AnyFunction, x);
```
Avaq: I tried different ways of testing:

1. `test([], Type, applyParameterizedType(x))` - This blows up if someone would provide, say `[x => true]` as `env`.
1. `test([], Type, x) || test([], Function([Type, Type]), x) || test([], Function([Type, Type, Type]), x)` - This effectively behaves the same as `AnyFunction`, because `test` does not apply `x` to determine its types.
1. This left me with the method seen above.

The decision, really, is between `1` and `3`, where `3` is overly permissive, and `1` blows up with an ugly error. Maybe to get the best of both worlds, I should combine them as:

```js
test([], Type, x) || test([], AnyFunction, x) && test([], Type, applyParameterizedType(x))
```

index.js
```diff
@@ -408,13 +408,16 @@
     return BinaryType(name, functionUrl(name), test, _1, _2);
   }
 
+  //  applyParameterizedType :: (Type... -> Type) -> Type
+  function applyParameterizedType(type) {
+    return typeof type === 'function' ?
+      type.apply(null, Z.map(K(Unknown), range(0, type.length))) :
+      type;
+  }
+
```
Avaq: See the reason for this diff in my other comment

index.js
davidchambers: This predicate looks fine to me. I think the ultimate solution is to require values such as `$.Array($.Unknown)` and `S.EitherType($.Unknown, $.Unknown)` in place of `$.Array` and `S.EitherType`. I hope VariadicType will be short-lived, so the predicate need not be perfect.

index.js
davidchambers: I think this should remain unchanged. It doesn't make sense to define `Array -> Array` or `Either -> Either`.

index.js
Avaq: So you vote for not combining `1` and `3`? In that case, shall I undo the diff that introduces `applyParameterizedType` (*singular*)?

index.js
Avaq: :zap:

index.js
davidchambers: I'd prefer not to include this refactoring to keep the diff as clear as possible. If you explained your reasons for making this its own function in another comment, I haven't been able to find the comment. ;)

Avaq: The reason was to enable the `test([], Type, applyParameterizedType(x))`-way of testing `VariadicType` membership.

index.js
Avaq: :zap:

Avaq:
:green_apple:

index.js
davidchambers: Ah, I see. I didn't make the connection, I'm sorry.

index.js
```diff
@@ -762,6 +762,14 @@
     typeEq('sanctuary-def/Type')
   );
 
+  var VariadicType = NullaryType(
```
davidchambers: Let's add a type signature along these lines:

```javascript
//  VariadicType :: Type | Type -> Type | Type -> Type -> Type | ...
```

The made-up nature of the signature will draw attention to our naughtiness.

Avaq: :zap:

davidchambers:
I'll close this in accordance with https://github.com/sanctuary-js/sanctuary-def/issues/123#issuecomment-282237805.

PR #122, opened 2017-02-23
Remove one of the env arguments from _determineActualTypes

Avaq:
Ha! I knew it! Is this damaging? Seems to be fine.

### DIFF
--- a/index.js
+++ b/index.js
@@ -841,7 +841,6 @@
   function _determineActualTypes(
     loose,          // :: Boolean
     env,            // :: Array Type
-    types,          // :: Array Type
     seen,           // :: Array Object
     values          // :: Array Any
   ) {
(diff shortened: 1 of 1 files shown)


davidchambers:
This is wonderful! Before I merge this, I'll look through the commit history in an effort to understand how we came to have both `env :: Array Type` and `types :: Array Type` and confirm that it's no longer necessary.

Avaq:
> an effort to understand how we came to have both

If memory serves me right, one of them used to be mutated.

davidchambers:
It looks as though this is the result of the performance optimization in #78. The key, I believe, is being able to provide `typeVarMap[typeVarName].types` rather than `env`, which makes a big difference when determining the types of a value such as `S.range(0, 1000)`.

Avaq:
I see! It seems you were using the extra `types` argument then, which you later didn't need anymore: https://github.com/sanctuary-js/sanctuary-def/commit/5d4032671d15456ba8f7eacea332fa9168eec13d#diff-168726dbe96b3ce427e7fedce31bb0bcL883

davidchambers:
Oh, good find! So you're suggesting that currently `env` and `types` are always identical? Let's check the call sites.

Recursive calls to `_determineActualTypes`:

  - `recur(loose, env, env, seen$, t.types.$1.extractor(value))`
  - `recur(loose, env, env, seen$, t.types.$1.extractor(value))`
  - `recur(loose, env, env, seen$, t.types.$2.extractor(value))`

Call from `determineActualTypesStrict`:

  - `_determineActualTypes(false, env, types, [], values)`

Call from `determineActualTypesLoose`:

  - `_determineActualTypes(true, env, types, [], values)`

All calls to `determineActualTypesStrict` and `determineActualTypesLoose` provide the same array for both `env` and `types`. This looks safe to merge! Shall we wait until #121 is merged before doing so? I'd like the patch release to contain as few changes as possible.

Avaq:
> I'd like the patch release to contain as few changes as possible.

Your call. :)

index.js
```diff
@@ -894,16 +893,16 @@
   }
 
   //  determineActualTypesStrict ::
-  //    (Array Type, Array Type, Array Any) -> Array Type
-  function determineActualTypesStrict(env, types, values) {
-    var types$ = _determineActualTypes(false, env, types, [], values);
+  //    (Array Type, Array Any) -> Array Type
```
davidchambers: It's no longer necessary to wrap this signature. :)

Avaq: :zap:

PR #132, opened 2017-04-08
sanctuary-type-classes@4.0.x

davidchambers:
https://github.com/sanctuary-js/sanctuary-type-classes/compare/v3.0.1...v4.0.0

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -85,9 +85,9 @@ function _Maybe(tag, value) {
 
 _Maybe['@@type'] = 'my-package/Maybe';
 
-_Maybe['fantasy-land/empty'] = function() { return Nothing; };
-
 _Maybe['fantasy-land/of'] = Just;
+
+_Maybe['fantasy-land/zero'] = function() { return Nothing; };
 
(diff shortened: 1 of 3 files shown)


test/index.js
```diff
@@ -85,10 +85,10 @@ function _Maybe(tag, value) {
 
 _Maybe['@@type'] = 'my-package/Maybe';
 
-_Maybe['fantasy-land/empty'] = function() { return Nothing; };
-
 _Maybe['fantasy-land/of'] = Just;
 
+_Maybe['fantasy-land/zero'] = function() { return Nothing; };
+
 _Maybe.prototype['fantasy-land/equals'] = function(other) {
```
gabejohnson: So `Maybe` will no longer implement `Monoid`?

davidchambers: This is not intended to be a complete definition. Once sanctuary-maybe is available we'll remove it entirely. For now, though, we need to implement methods to allow `filter` (defined a few thousand lines further down) to operate on `Maybe a` values. Currently this requires `fantasy-land/empty`, but with upgrade it requires `fantasy-land/zero` instead.

PR #134, opened 2017-04-13
types: add Symbol

### LINKED ISSUE #133: Add $.Symbol
In https://github.com/sanctuary-js/sanctuary/issues/381 I mentioned that `R.pluck` works with Arrays and Number indices as opposed to just StrMaps and String keys.

The signature of `S.pluck` _could_ be:
```hs
type Natural = PositiveInteger | 0

type AccessibleKey = String | Natural | Symbol

pluck :: (Accessible a, Functor f) => AccessibleKey -> f a -> f b
```

This would require `$.Symbol` (and possibly `$.Natural`).

gabejohnson:
Closes #133

### DIFF
--- a/index.js
+++ b/index.js
@@ -713,6 +713,11 @@
   //. Type comprising every primitive String value.
   var String_ = NullaryTypeWithUrl('String', typeofEq('string'));
 
+  //# Symbol :: Type
+  //.
+  //. Type comprising every Symbol value.
+  var Symbol_ = NullaryTypeWithUrl('Symbol', typeofEq('symbol'));
+
   //# Type :: Type
   //.
   //. Type comprising every `Type` value.
@@ -779,6 +784,7 @@
   //.   - <code><a href="#RegExp">RegExp</a></code>
   //.   - <code><a href="#StrMap">StrMap</a>(<a href="#Unknown">Unknown</a>)</code>
   //.   - <code><a href="#String">String</a></code>
+  //.   - <code><a href="#Symbol">Symbol</a></code>
   //.   - <code><a href="#Undefined">Undefined</a></code>
   var env = [
     AnyFunction,
@@ -793,6 +799,7 @@
     RegExp_,
     StrMap(Unknown),
     String_,
+    Symbol_,
     Undefined
   ];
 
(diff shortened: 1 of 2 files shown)


gabejohnson:
@davidchambers will this change require adding Symbol to sanctuary-type-classes as well?

davidchambers:
> @davidchambers will this change require adding Symbol to sanctuary-type-classes as well?

I don't think so. What do you have in mind?

index.js
```diff
@@ -713,6 +713,12 @@
   //. Type comprising every primitive String value.
   var String_ = NullaryTypeWithUrl('String', typeofEq('string'));
 
+
```
davidchambers: Nit: Unwanted line.

index.js
```diff
@@ -713,6 +713,12 @@
...
+  //# Symbol :: Type
+  //.
+  //. Type comprising every primitive Symbol value.
```
davidchambers: Is it necessary to use "primitive" here?

index.js
```diff
@@ -713,6 +713,12 @@
...
+  var Symbol_ = NullaryTypeWithUrl('Symbol', typeofEq('symbol'));
```
davidchambers: Is there a reason to favour `typeofEq('symbol')` over `typeEq('Symbol')`, or are the expressions equivalent?

index.js
```diff
@@ -793,6 +800,7 @@
     RegExp_,
     StrMap(Unknown),
     String_,
+    Symbol,
```
davidchambers: `s/Symbol/Symbol_/`

index.js
gabejohnson: An artifact of copying. There is no boxed version of `Symbol` thus no need to differentiate it.

Though it is a primitive.

index.js
gabejohnson: They should be equivalent. I would think the `typeofEq` would be slightly faster. I don't know if there are any benchmarks though.

gabejohnson:
Updated

index.js
davidchambers: :thumbsup:

PR #145, opened 2017-05-01
types: define NonEmpty type constructor

davidchambers:
I'll explain how I arrived at this pull request.

1.  I wrote something like this, forgetting to specify the types:

    ```javascript
    const f = def('f', {}, [], _f);
    ```

2.  I ran the code and got an unhelpful error message:

    ```
    RangeError: Invalid array length
    ```

3.  I set out to add a length check and a custom error message for an empty type array.

4.  I remembered that we have an elegant way to avoid ad hoc type checking. :D

5.  I defined `NonEmptyArray` for internal use to improve the error message.

6.  I decided to export `NonEmptyArray` (and `NonEmptyString`, for symmetry).

### DIFF
--- a/index.js
+++ b/index.js
@@ -569,6 +569,23 @@
     function(x) { return Number_._test(x) && x < 0; }
   );
 
+  //# NonEmpty :: Type -> Type
+  //.
+  //. Constructor for non-empty types. `$.NonEmpty($.String)`, for example, is
+  //. the type comprising every [`String`][] value except `''`.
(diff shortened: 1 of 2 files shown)


safareli:
I think you can define one `NonEmpty(Type)`  for any Type which is `Eq` and `Monoid`

davidchambers:
Wonderful suggestion, Irakli! Please have a look at the updated patch. :)

PR #147, opened 2017-05-01
perform type checking when a "returned" function is applied

### LINKED ISSUE #146: type checking should be performed when a "returned" function is applied
In sanctuary-js/sanctuary#388 we'd like to create functions of type

```haskell
:: Ord a => a -> (a -> Boolean)
```

rather than

```haskell
:: Ord a => a -> a -> Boolean
```

in order to prevent misuse. This can be done via `[a, $.Function([a, $.Boolean])]`. The problem is that no type checking is performed when one of the resulting `$.Function([a, $.Boolean])` functions is applied.

Here's a simplified test case:

```javascript
const $ = require('sanctuary-def');

const def = $.create({checkTypes: true, env: $.env});

//    lt :: Number -> (Number -> Number)
const lt =
def('lt',
    {},
    [$.Number, $.Function([$.Number, $.Number])],
    y => x => x < y);

lt(123)('abc');
```

`lt(123)('abc')` should be a type error, but currently is not.

davidchambers:
Closes #146

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -235,7 +235,7 @@ describe('create', function() {
            TypeError,
            'Invalid value\n' +
            '\n' +
-           'create :: { checkTypes :: Boolean, env :: Array Any } -> ((String, StrMap (Array TypeClass), Array Type, Function) -> Function)\n' +
+           'create :: { checkTypes :: Boolean, env :: Array Any } -> Function\n' +
            '          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n' +
            '                               1\n' +
            '\n' +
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -11,7 +11,7 @@
     "test": "make test lint"
   },
   "dependencies": {
-    "sanctuary-type-classes": "4.0.x",
+    "sanctuary-type-classes": "5.0.x",
```
davidchambers: This should happen in a separate pull request, before this pull request is merged.

test/index.js
```diff
@@ -2942,7 +2969,7 @@ describe('UnaryType', function() {
   it('is a quaternary function', function() {
     eq(typeof $.UnaryType, 'function');
     eq($.UnaryType.length, 4);
-    eq($.UnaryType.toString(), 'UnaryType :: String -> String -> (Any -> Boolean) -> (t a -> Array a) -> (Type -> Type)');
+    eq($.UnaryType.toString(), 'UnaryType :: String -> String -> (Any -> Boolean) -> (t a -> Array a) -> Function');
```
davidchambers: This is the result of changing the return type from `Function_([Type, Type])` to `AnyFunction`. Why make this change? `$.UnaryType` returns a function which *itself* performs type checking. Leaving type checking to the inner function results in clearer error messages. Consider, for example, the treatment of `$.Array(0)`:

```diff
-UnaryType :: String -> String -> (Any -> Boolean) -> (t a -> Array a) -> (Type -> Type)
-                                                                          ^^^^
-                                                                           1
+Array :: Type -> Type
+         ^^^^
+          1
```

PR #153, opened 2017-05-28
sanctuary-type-classes@6.0.x

davidchambers:
<https://github.com/sanctuary-js/sanctuary-type-classes/compare/v5.1.0...v6.0.0>

Thanks to sanctuary-js/sanctuary-type-classes#57 we can now link to a type class's documentation from the error message for a type-class constraint violation.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1919,6 +1919,7 @@
   //. //    Semigroup :: TypeClass
   //. const Semigroup = Z.TypeClass(
   //.   'my-package/Semigroup',
+  //.   'http://example.com/my-package#Semigroup',
   //.   [],
   //.   x => x != null && typeof x.concat === 'function'
   //. );
@@ -1940,6 +1941,8 @@
   //. //   1)  {} :: Object, StrMap ???
   //. //
   //. //   ‘concat’ requires ‘a’ to satisfy the Semigroup type-class constraint; the value at position 1 does not.
+  //. //
+  //. //   See http://example.com/my-package#Semigroup for information about the my-package/Semigroup type class.
   //.
   //. concat(null, null);
   //. // ! TypeError: Type-class constraint violation
@@ -1951,6 +1954,8 @@
   //. //   1)  null :: Null
   //. //
   //. //   ‘concat’ requires ‘a’ to satisfy the Semigroup type-class constraint; the value at position 1 does not.
+  //. //
+  //. //   See http://example.com/my-package#Semigroup for information about the my-package/Semigroup type class.
   //. ```
   //.
   //. Multiple constraints may be placed on a type variable by including
@@ -2179,6 +2184,13 @@
     };
   }
 
+  //  see :: (String, { name :: String, url :: String }) -> String
+  function see(label, record) {
+    return record.url &&
+           '\nSee ' + record.url +
+           ' for information about the ' + record.name + ' ' + label + '.\n';
+  }
+
   //  typeClassConstraintViolation :: ... -> Error
   function typeClassConstraintViolation(
     env,            // :: Array Type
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -2179,6 +2184,13 @@
     };
   }
 
+  //  see :: (String, { name :: String, url :: String }) -> String
+  function see(label, record) {
+    return record.url &&
```
Avaq: If `record.url` is undefined, this function would return `undefined`, right, violating the type signature? I'm looking into it because of fluture-js/fluture-sanctuary-types#5. I believe if sanctuary-def receives type classes from an older version of sanctuary-type-classes, its errors will now include: `"\n undefined"`.

Alternatively, the type class `@@type` property has changed (`@version`-part?) and sanctuary-def will throw an error if getting an older type class ... in which case older versions of sanctuary-type-classes no longer work with newer versions of sanctuary-def, making https://github.com/fluture-js/Fluture/commit/37822c5068f3c207212fd934735e67af40cf19f3 an indirect breaking change, forcing me to bump the version of fluture-sanctuary-types.

Avaq:
```sh
$ npm i sanctuary-type-classes@4
$ npm i sanctuary-def@latest
node
```

```js
var Z = require('sanctuary-type-classes')
var $ = require('sanctuary-def')
var def = $.create({checkTypes: true, env: $.env})
var f = $.UnaryTypeVariable('f')
var a = $.TypeVariable('a')
var b = $.TypeVariable('b')
var map = def('map', {f: [Z.Functor]}, [$.Function([a, b]), f(a), f(b)], Z.map)

map(x => x + 1, 1)
```

```text
TypeError: Type-class constraint violation

map :: Functor f => (a -> b) -> f a -> f b
       ^^^^^^^^^                ^^^
                                 1

1)  1 :: Number

‘map’ requires ‘f’ to satisfy the Functor type-class constraint; the value at position 1 does not.
undefined
```

----

So indeed. It's not a breaking change to use an outdated type-classes library, but it *is* going to lead to a trailing "undefined" at the end of your error messages.

index.js
Avaq:
```sh
$ npm i sanctuary-type-classes@4
$ npm i sanctuary-def@latest
node
```

```js
var Z = require('sanctuary-type-classes')
var $ = require('sanctuary-def')
var def = $.create({checkTypes: true, env: $.env})
var f = $.UnaryTypeVariable('f')
var a = $.TypeVariable('a')
var b = $.TypeVariable('b')
var map = def('map', {f: [Z.Functor]}, [$.Function([a, b]), f(a), f(b)], Z.map)

map(x => x + 1, 1)
```

```text
TypeError: Type-class constraint violation

map :: Functor f => (a -> b) -> f a -> f b
       ^^^^^^^^^                ^^^
                                 1

1)  1 :: Number

‘map’ requires ‘f’ to satisfy the Functor type-class constraint; the value at position 1 does not.
undefined
```

----

So indeed. It's not a breaking change to use an outdated type-classes library, but it *is* going to lead to a trailing "undefined" at the end of your error messages.

Avaq: I believe that we can release a `patch` to sanctuary-def in order to combat this issue, and allow users to enjoy using their outdated fluture or other sanctuary-type-classes based library with sanctuary-def without trailing `undefined`s in their error messages. ;)

davidchambers: I'll freely admit that I'm not in the habit of considering how code will interoperate with old versions of related packages. Tricky!

Your fix sounds good to me. Would you like to submit a pull request?

Avaq: I think very few library authors do. I've considered this matter and reached the conclusion that it may be better that if your dependency exposes anything to the user, it should be a peer dependency rather than a regular dependency. And if your library works with input from another library, you should specify its version range using an [optional dependency](https://docs.npmjs.com/files/package.json#optionaldependencies). But I'm still reflecting upon these thoughts.

Avaq: I'll submit a PR shortly.

PR #156, opened 2017-05-30
types: define Accessible

davidchambers:


davidchambers:
I think Accessible is more useful as a type class, which is how it is defined in Sanctuary.

PR #157, opened 2017-06-01
Allow see() to handle records without URLs

Avaq:
See https://github.com/sanctuary-js/sanctuary-def/pull/153#discussion_r119550590

Commit body:

> This ensures that records like the ones created by older
versions of sanctuary-type-classes are not being treated
like garbage. As a result we don't see "undefined" in
error messages generated using records with missing
"url" properties.

### DIFF
--- a/index.js
+++ b/index.js
@@ -2191,9 +2191,10 @@
     };
   }
 
-  //  see :: (String, { name :: String, url :: String }) -> String
+  //  see :: (String, { name :: String, url :: String? }) -> String
   function see(label, record) {
-    return record.url &&
+    return record.url == null || record.url === '' ?
+           '' :
            '\nSee ' + record.url +
            ' for information about the ' + record.name + ' ' + label + '.\n';
   }


index.js
```diff
@@ -2191,11 +2191,12 @@
     };
   }
 
-  //  see :: (String, { name :: String, url :: String }) -> String
+  //  see :: (String, { name :: String, url :: Nullable String }) -> String
```
davidchambers: Let's use `String?` rather than `Nullable String` to reflect the fact that the field may not be present.

index.js
```diff
@@ -2191,11 +2191,12 @@
...
   function see(label, record) {
-    return record.url &&
+    return record.url ?
```
davidchambers: I'd like to make the checks for `undefined` and `''` explicit:

```javascript
return record.url == null || record.url === '' ?
  '' :
  '\nSee ' + ... + '.\n';
```

PR #160, opened 2017-06-22
fix handling of constraint violations of nested type variables

davidchambers:
Fixes sanctuary-js/sanctuary#412

### DIFF
--- a/index.js
+++ b/index.js
@@ -280,6 +280,13 @@
     return result;
   }
 
+  //  singleton :: (String, a) -> StrMap a
+  function singleton(k, v) {
+    var result = {};
+    result[k] = v;
+    return result;
+  }
+
   //  strRepeat :: (String, Integer) -> String
   function strRepeat(s, times) {
     return Array(times + 1).join(s);
@@ -903,7 +910,6 @@
 
   //  _determineActualTypes :: ... -> Array Type
   function _determineActualTypes(
-    loose,          // :: Boolean
     env,            // :: Array Type
     seen,           // :: Array Object
     values          // :: Array Any
@@ -927,15 +933,15 @@
             [] :
           t.type === UNARY ?
             Z.map(fromUnaryType(t),
-                  recur(loose, env, seen$, t.types.$1.extractor(value))) :
+                  recur(env, seen$, t.types.$1.extractor(value))) :
           t.type === BINARY ?
             xprod(
               t,
               t.types.$1.type.type === UNKNOWN ?
-                recur(loose, env, seen$, t.types.$1.extractor(value)) :
+                recur(env, seen$, t.types.$1.extractor(value)) :
                 [t.types.$1.type],
               t.types.$2.type.type === UNKNOWN ?
-                recur(loose, env, seen$, t.types.$2.extractor(value)) :
+                recur(env, seen$, t.types.$2.extractor(value)) :
                 [t.types.$2.type]
             ) :
           // else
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1171,14 +1164,23 @@
...
+                                                                typeInfo,
+                                                                index,
+                                                                propPath$,
+                                                                valuesByPath);
+                            }()) :
+                            invalidValue(env, typeInfo, index, propPath$, x);
                         });
                       }, e);
-                    }, Right(r), innerValues)
+                    }, e, innerValues)
```
davidchambers: This was a type error that went undetected for a long time. `r` was in fact a value of type `Either a b`, so `Right(r)` produced a value of type `Either a (Either a b)` which was incorrect. The correct expression was simply `r`.

I renamed the parameter to hint that it's of type `Either a b`.

index.js
```diff
@@ -939,26 +938,27 @@
...
-    return types.filter(function(t) {
-      return t.type !== INCONSISTENT && t.type !== UNKNOWN;
-    });
+  //  isConsistent :: Type -> Boolean
+  function isConsistent(t) {
+    return t.type === UNARY   ? isConsistent(t.types.$1.type) :
+           t.type === BINARY  ? isConsistent(t.types.$1.type) &&
+                                isConsistent(t.types.$2.type) :
+           /* else */           t.type !== INCONSISTENT;
   }
```
davidchambers: This function is an island of sanity in a sea of complicated code. I'm pleased that we're no longer treating `$.Unknown` as inconsistent. :)

index.js
```diff
@@ -939,26 +938,27 @@
...
+           .filter(isConsistent);
   }
 
   //  determineActualTypesLoose :: (Array Type, Array Any) -> Array Type
   function determineActualTypesLoose(env, values) {
-    var types$ = _determineActualTypes(true, env, [], values);
-    return rejectInconsistent(types$);
+    return _determineActualTypes(env, [], values)
+           .filter(function(t) { return t.type !== INCONSISTENT; });
   }
```
davidchambers: The difference between the `*Strict` and `*Loose` functions is that the latter permits inconsistent types such as `Array ???` in order to mention these in error messages.

index.js
```diff
@@ -1059,7 +1052,7 @@
             return function(propPath) {
               var indexedPropPath = Z.concat([index], propPath);
               return function(s) {
-                if (t.type === VARIABLE) {
+                if (t.type === VARIABLE && isEmpty(t.keys)) {
```
davidchambers: This enables us to underline the `a` in `f a`.

index.js
```diff
@@ -1171,14 +1164,23 @@
...
+                                                              propPath$));
+                              var valuesByPath = {};
+                              valuesByPath[k] = [x];
+                              return typeVarConstraintViolation(env,
+                                                                typeInfo,
+                                                                index,
+                                                                propPath$,
+                                                                valuesByPath);
+                            }()) :
+                            invalidValue(env, typeInfo, index, propPath$, x);
```
davidchambers: This change doesn't feel very good. :\

test/index.js
```diff
@@ -2989,6 +3003,18 @@ describe('def', function() {
...
+           TypeError,
+           'Type-variable constraint violation\n' +
+           '\n' +
+           'bimap :: Bifunctor f => (a -> b) -> (c -> d) -> f a c -> f b d\n' +
+           '                                                  ^\n' +
+           '                                                  1\n' +
+           '\n' +
+           '1)  ["foo", true, 42] :: Array ???\n' +
+           '\n' +
+           'Since there is no type of which all the above values are members, the type-variable constraint has been violated.\n');
```
davidchambers: How do these messages look to you, @Avaq?

Avaq:
I'll need to dive in to review these changes. I will get to it.

davidchambers:
Thanks, Aldwin. The forecast is for showers in your part of the world tomorrow. :umbrella:

test/index.js
Avaq: Shouldn't the second message say the following?

```text
1)  ["foo", true, 42] :: Array ???
    null :: Null

Since there is no type of which all the above values are members ...
```

Avaq: Ah, no. Never mind. For some reason I thought for a second Tuple needs both inner values to belong to the same type.

It still seems confusing to me though, that the output shows that we found a type for `["foo", true, 42]` (namely `Array ???`), but then we say that there's no types for that value.

Avaq: If I implement `showType` as such:

```js
  //  showType :: Type -> String
  function showType(t) {
    return isConsistent(t) ?
           unless(t.type === FUNCTION || t.type === RECORD || isEmpty(t.keys),
                  stripOutermostParens,
                  String(t)) :
           '(inconsistent)';
  }
```

Then any inconsistent type will simply be rendered as `(inconsistent)`, this removes the confusion between values for which the type has not been determined (such as the inner type of a `Nothing`), and values determined to be inconsistent and therefore members of no types. For me this is clearer, but I also realize now how minor the issue is.

Another approach could be to change `always('???')` to `always('(inconsistent)')` as the render function of Inconsistent, which would make this type render as `Array (inconsistent)`. This also alleviates some confusion, but then we leave the fact that an array of inconsistent types is "not an actual type" implied.

Avaq: Playing around, I think I prefer the second approach.

davidchambers: #163

PR #162, opened 2017-06-26
Type refinements

### LINKED ISSUE #128: "strict" record types
Record types permit additional fields. For example:

```javascript

$.test([], Point, {x: 0, y: 0});
// => true

$.test([], Point, {x: 0, y: 0, radius: 5});
// => true
```

One may wish to define a "strict" record type instead. One may do so using [`$.NullaryType`][1]:

```javascript
const StrictPoint = $.NullaryType(
  'my-package/StrictPoint',
  '',
  x => $.test([], Point, x) && Object.keys(x).length === Point.keys.length
);

$.test([], StrictPoint, {x: 0, y: 0});
// => true

$.test([], StrictPoint, {x: 0, y: 0, radius: 5});
// => false
```

We could provide `$.StrictRecordType`, or have [`$.RecordType`][2] take an additional argument to specify whether additional fields are permitted. I'm not sure we should do so, though, given the solution above.

This issue serves both to document the existing solution, and to provide a place to discuss the pros and cons of supporting "strict" record types more directly.

[1]: https://github.com/sanctuary-js/sanctuary-def#NullaryType
[2]: https://github.com/sanctuary-js/sanctuary-def#RecordType

### LINKED ISSUE #141: Prototypical Record Types?
Currently `$.RecordType` checks a value via `Object.keys`.  `Object.keys` will not pick up prototypical properties and methods.

I think it'd be pretty helpful to be able to define types that include prototypes.

E.g.

```
const $Promise = 
  $.RecordType({
    then: $.AnyFunction
    ,catch: $.AnyFunction
  })
```

Maybe `$Promise` isn't the best example, we'd probably want to use a BinaryType in that case.  But even internally we could use the above in the validator for a BinaryType version for brevity.

Avaq:
Refinement support:

- [x] NullaryType
- [x] UnaryType
- [x] BinaryType
- [x] RecordType
- [ ] EnumType

Helpers:

- [x] Strict
- [ ] NonEmpty

Fixes #128 
Fixes #141 
Contributes to sanctuary-js/sanctuary#384

### DIFF
--- a/index.js
+++ b/index.js
@@ -2602,8 +2602,12 @@
                def);
   }
 
-  var Options = RecordType(Object_, {checkTypes: Boolean_, env: Array_(Type)});
-  var create = def('create', {}, [Options, AnyFunction], _create);
+  var create =
+  def('create',
+      {},
+      [RecordType(Object_, {checkTypes: Boolean_, env: Array_(Type)}),
+       AnyFunction],
+      _create);
 
   //  fromUncheckedUnaryType :: (Type -> Type) -> (Type -> Type)
   function fromUncheckedUnaryType(typeConstructor) {
--- a/index.js
+++ b/index.js
@@ -466,15 +466,6 @@
   Any.test = K(true);
   Any.validate = Right;
 
-  //# None :: Type
-  //.
-  //. Type with no members.
-  var None = NullaryTypeWithUrl('sanctuary-def/None', Any, K(false));
-
-  //  Inconsistent :: Type
-  var Inconsistent =
-  new _Type(INCONSISTENT, '', '', always2('???'), None, K(false), [], {});
-
   //# AnyFunction :: Type
   //.
   //. Type comprising every Function value.
@@ -500,15 +491,34 @@
   //. Type comprising every Date value.
   var Date_ = NullaryTypeWithUrl('Date', Any, typeEq('Date'));
 
+  //# ValidDate :: Type
+  //.
+  //. Type comprising every [`Date`][] value except `new Date(NaN)`.
+  var ValidDate = NullaryTypeWithUrl(
+    'sanctuary-def/ValidDate',
+    Date_,
+    function(x) { return !isNaN(x.valueOf()); }
+  );
+
   //# Error :: Type
   //.
   //. Type comprising every Error value, including values of more specific
   //. constructors such as [`SyntaxError`][] and [`TypeError`][].
   var Error_ = NullaryTypeWithUrl('Error', Any, typeEq('Error'));
 
+  //  Inconsistent :: Type
+  var Inconsistent =
+  new _Type(INCONSISTENT, '', '', always2('???'), Any, K(false), [], {});
+
+  //# Null :: Type
+  //.
+  //. Type whose sole member is `null`.
+  var Null = NullaryTypeWithUrl('Null', Any, typeEq('Null'));
+
   //# Number :: Type
   //.
-  //. Type comprising every primitive Number value (including `NaN`).
+  //. Type comprising every primitive Number value
+  //. (including `NaN` and `Infinity`).
   var Number_ = NullaryTypeWithUrl('Number', Any, typeofEq('number'));
 
   //# ValidNumber :: Type
@@ -530,62 +540,31 @@
     isFinite
   );
 
-  //# Function :: Array Type -> Type
+  //# NegativeFiniteNumber :: Type
   //.
-  //. Constructor for Function types.
-  //.
-  //. Examples:
-  //.
-  //.   - `$.Function([$.Date, $.String])` represents the `Date -> String`
-  //.     type; and
-  //.   - `$.Function([a, b, a])` represents the `(a, b) -> a` type.
-  function Function_(types) {
-    function format(outer, inner) {
-      var xs = types.map(function(t, idx) {
-        return unless(t.type === RECORD || isEmpty(t.keys),
-                      stripOutermostParens,
-                      inner('$' + String(idx + 1))(String(t)));
-      });
-      var parenthesize = wrap(outer('('))(outer(')'));
-      return parenthesize(unless(types.length === 2,
-                                 parenthesize,
-                                 init(xs).join(outer(', '))) +
-                          outer(' -> ') +
-                          last(xs));
-    }
+  //. Type comprising every [`FiniteNumber`][] less than zero.
+  var NegativeFiniteNumber = NullaryTypeWithUrl(
+    'sanctuary-def/NegativeFiniteNumber',
+    FiniteNumber,
+    function(x) { return x < 0; }
+  );
 
-    var $keys = [];
-    var $types = {};
-    types.forEach(function(t, idx) {
-      var k = '$' + String(idx + 1);
-      $keys.push(k);
-      $types[k] = {extractor: K([]), type: t};
-    });
-
-    return new _Type(FUNCTION,
-                     '',
-                     '',
-                     format,
-                     AnyFunction,
-                     K(true),
-                     $keys,
-                     $types);
-  }
-
-  //# RegExp :: Type
+  //# NonZeroFiniteNumber :: Type
   //.
-  //. Type comprising every RegExp value.
-  var RegExp_ = NullaryTypeWithUrl('RegExp', Any, typeEq('RegExp'));
+  //. Type comprising every [`FiniteNumber`][] value except `0` and `-0`.
+  var NonZeroFiniteNumber = NullaryTypeWithUrl(
+    'sanctuary-def/NonZeroFiniteNumber',
+    FiniteNumber,
+    function(x) { return x !== 0; }
+  );
 
-  //# GlobalRegExp :: Type
+  //# PositiveFiniteNumber :: Type
   //.
-  //. Type comprising every [`RegExp`][] value whose `global` flag is `true`.
-  //.
-  //. See also [`NonGlobalRegExp`][].
-  var GlobalRegExp = NullaryTypeWithUrl(
-    'sanctuary-def/GlobalRegExp',
-    RegExp_,
-    function(x) { return x.global; }
+  //. Type comprising every [`FiniteNumber`][] value greater than zero.
+  var PositiveFiniteNumber = NullaryTypeWithUrl(
+    'sanctuary-def/PositiveFiniteNumber',
+    FiniteNumber,
+    function(x) { return x > 0; }
   );
 
   //# Integer :: Type
@@ -602,15 +581,6 @@
     }
   );
 
-  //# NegativeFiniteNumber :: Type
-  //.
-  //. Type comprising every [`FiniteNumber`][] value less than zero.
-  var NegativeFiniteNumber = NullaryTypeWithUrl(
-    'sanctuary-def/NegativeFiniteNumber',
-    FiniteNumber,
-    function(x) { return x < 0; }
-  );
-
   //# NegativeInteger :: Type
   //.
   //. Type comprising every [`Integer`][] value less than zero.
@@ -618,52 +588,6 @@
     'sanctuary-def/NegativeInteger',
     Integer,
     function(x) { return x < 0; }
-  );
-
-  //# NegativeNumber :: Type
-  //.
-  //. Type comprising every [`Number`][] value less than zero.
-  var NegativeNumber = NullaryTypeWithUrl(
-    'sanctuary-def/NegativeNumber',
-    Number_,
-    function(x) { return x < 0; }
-  );
-
-  //# NonEmpty :: Type -> Type
-  //.
-  //. Constructor for non-empty types. `$.NonEmpty($.String)`, for example, is
-  //. the type comprising every [`String`][] value except `''`.
-  //.
-  //. The given type must satisfy the [Monoid][] and [Setoid][] specifications.
-  var NonEmpty = UnaryType(
-    'sanctuary-def/NonEmpty',
-    functionUrl('NonEmpty'),
-    function(x) {
-      return Z.Monoid.test(x) &&
-             Z.Setoid.test(x) &&
-             !Z.equals(x, Z.empty(x.constructor));
-    },
-    function(monoid) { return [monoid]; }
-  );
-
-  //# NonGlobalRegExp :: Type
-  //.
-  //. Type comprising every [`RegExp`][] value whose `global` flag is `false`.
-  //.
-  //. See also [`GlobalRegExp`][].
-  var NonGlobalRegExp = NullaryTypeWithUrl(
-    'sanctuary-def/NonGlobalRegExp',
-    RegExp_,
-    function(x) { return !x.global; }
-  );
-
-  //# NonZeroFiniteNumber :: Type
-  //.
-  //. Type comprising every [`FiniteNumber`][] value except `0` and `-0`.
-  var NonZeroFiniteNumber = NullaryTypeWithUrl(
-    'sanctuary-def/NonZeroFiniteNumber',
-    FiniteNumber,
-    function(x) { return x !== 0; }
   );
 
   //# NonZeroInteger :: Type
@@ -675,6 +599,24 @@
     function(x) { return x !== 0; }
   );
 
+  //# PositiveInteger :: Type
+  //.
+  //. Type comprising every [`Integer`][] value greater than zero.
+  var PositiveInteger = NullaryTypeWithUrl(
+    'sanctuary-def/PositiveInteger',
+    Integer,
+    function(x) { return x > 0; }
+  );
+
+  //# NegativeNumber :: Type
+  //.
+  //. Type comprising every [`ValidNumber`][] value less than zero.
+  var NegativeNumber = NullaryTypeWithUrl(
+    'sanctuary-def/NegativeNumber',
+    ValidNumber,
+    function(x) { return x < 0; }
+  );
+
   //# NonZeroValidNumber :: Type
   //.
   //. Type comprising every [`ValidNumber`][] value except `0` and `-0`.
@@ -684,21 +626,13 @@
     function(x) { return x !== 0; }
   );
 
-  //# Null :: Type
+  //# PositiveNumber :: Type
   //.
-  //. Type whose sole member is `null`.
-  var Null = NullaryTypeWithUrl('Null', Any, typeEq('Null'));
-
-  //# Nullable :: Type -> Type
-  //.
-  //. Constructor for types which include `null` as a member.
-  var Nullable = UnaryTypeWithUrl(
-    'sanctuary-def/Nullable',
-    K(true),
-    function(nullable) {
-      // eslint-disable-next-line eqeqeq
-      return nullable === null ? [] : [nullable];
-    }
+  //. Type comprising every [`ValidNumber`][] value greater than zero.
+  var PositiveNumber = NullaryTypeWithUrl(
+    'sanctuary-def/PositiveNumber',
+    ValidNumber,
+    function(x) { return x > 0; }
   );
 
   //# Object :: Type
@@ -711,44 +645,6 @@
   //.   - the `new` operator in conjunction with `Object` or a custom
   //.     constructor function.
   var Object_ = NullaryTypeWithUrl('Object', Any, typeEq('Object'));
-
-  //# Pair :: Type -> Type -> Type
-  //.
-  //. Constructor for tuple types of length 2. Arrays are said to represent
-  //. tuples. `['foo', 42]` is a member of `Pair String Number`.
-  var Pair = BinaryTypeWithUrl(
-    'sanctuary-def/Pair',
-    function(x) { return typeEq('Array')(x) && x.length === 2; },
-    function(pair) { return [pair[0]]; },
-    function(pair) { return [pair[1]]; }
-  );
-
-  //# PositiveFiniteNumber :: Type
-  //.
-  //. Type comprising every [`FiniteNumber`][] value greater than zero.
-  var PositiveFiniteNumber = NullaryTypeWithUrl(
-    'sanctuary-def/PositiveFiniteNumber',
-    FiniteNumber,
-    function(x) { return x > 0; }
-  );
-
-  //# PositiveInteger :: Type
-  //.
-  //. Type comprising every [`Integer`][] value greater than zero.
-  var PositiveInteger = NullaryTypeWithUrl(
-    'sanctuary-def/PositiveInteger',
-    Integer,
-    function(x) { return x > 0; }
-  );
-
-  //# PositiveNumber :: Type
-  //.
-  //. Type comprising every [`Number`][] value greater than zero.
-  var PositiveNumber = NullaryTypeWithUrl(
-    'sanctuary-def/PositiveNumber',
-    Number_,
-    function(x) { return x > 0; }
-  );
 
   //# RegexFlags :: Type
   //.
@@ -767,18 +663,31 @@
     ['', 'g', 'i', 'm', 'gi', 'gm', 'im', 'gim']
   );
 
-  //# StrMap :: Type -> Type
+  //# RegExp :: Type
   //.
-  //. Constructor for homogeneous Object types.
+  //. Type comprising every RegExp value.
+  var RegExp_ = NullaryTypeWithUrl('RegExp', Any, typeEq('RegExp'));
+
+  //# GlobalRegExp :: Type
   //.
-  //. `{foo: 1, bar: 2, baz: 3}`, for example, is a member of `StrMap Number`;
-  //. `{foo: 1, bar: 2, baz: 'XXX'}` is not.
-  var StrMap = UnaryTypeWithUrl(
-    'sanctuary-def/StrMap',
-    Object_.test.bind(Object_),
-    function(strMap) {
-      return Z.reduce(function(xs, x) { return xs.concat([x]); }, [], strMap);
-    }
+  //. Type comprising every [`RegExp`][] value whose `global` flag is `true`.
+  //.
+  //. See also [`NonGlobalRegExp`][].
+  var GlobalRegExp = NullaryTypeWithUrl(
+    'sanctuary-def/GlobalRegExp',
+    RegExp_,
+    function(x) { return x.global; }
+  );
+
+  //# NonGlobalRegExp :: Type
+  //.
+  //. Type comprising every [`RegExp`][] value whose `global` flag is `false`.
+  //.
+  //. See also [`GlobalRegExp`][].
+  var NonGlobalRegExp = NullaryTypeWithUrl(
+    'sanctuary-def/NonGlobalRegExp',
+    RegExp_,
+    function(x) { return !x.global; }
   );
 
   //# String :: Type
@@ -829,13 +738,100 @@
   var Unknown =
   new _Type(UNKNOWN, '', '', always2('???'), Any, K(true), [], {});
 
-  //# ValidDate :: Type
+  //# Function :: Array Type -> Type
   //.
-  //. Type comprising every [`Date`][] value except `new Date(NaN)`.
-  var ValidDate = NullaryTypeWithUrl(
-    'sanctuary-def/ValidDate',
-    Date_,
-    function(x) { return !isNaN(x.valueOf()); }
+  //. Constructor for Function types.
+  //.
+  //. Examples:
+  //.
+  //.   - `$.Function([$.Date, $.String])` represents the `Date -> String`
+  //.     type; and
+  //.   - `$.Function([a, b, a])` represents the `(a, b) -> a` type.
+  function Function_(types) {
+    function format(outer, inner) {
+      var xs = types.map(function(t, idx) {
+        return unless(t.type === RECORD || isEmpty(t.keys),
+                      stripOutermostParens,
+                      inner('$' + String(idx + 1))(String(t)));
+      });
+      var parenthesize = wrap(outer('('))(outer(')'));
+      return parenthesize(unless(types.length === 2,
+                                 parenthesize,
+                                 init(xs).join(outer(', '))) +
+                          outer(' -> ') +
+                          last(xs));
+    }
+
+    var $keys = [];
+    var $types = {};
+    types.forEach(function(t, idx) {
+      var k = '$' + String(idx + 1);
+      $keys.push(k);
+      $types[k] = {extractor: K([]), type: t};
+    });
+
+    return new _Type(FUNCTION,
+                     '',
+                     '',
+                     format,
+                     AnyFunction,
+                     K(true),
+                     $keys,
+                     $types);
+  }
+
+  //# NonEmpty :: Type -> Type
+  //.
+  //. Constructor for non-empty types. `$.NonEmpty($.String)`, for example, is
+  //. the type comprising every [`String`][] value except `''`.
+  //.
+  //. The given type must satisfy the [Monoid][] and [Setoid][] specifications.
+  var NonEmpty = UnaryType(
+    'sanctuary-def/NonEmpty',
+    functionUrl('NonEmpty'),
+    function(x) {
+      return Z.Monoid.test(x) &&
+             Z.Setoid.test(x) &&
+             !Z.equals(x, Z.empty(x.constructor));
+    },
+    function(monoid) { return [monoid]; }
+  );
+
+  //# Nullable :: Type -> Type
+  //.
+  //. Constructor for types which include `null` as a member.
+  var Nullable = UnaryTypeWithUrl(
+    'sanctuary-def/Nullable',
+    K(true),
+    function(nullable) {
+      // eslint-disable-next-line eqeqeq
+      return nullable === null ? [] : [nullable];
+    }
+  );
+
+  //# Pair :: Type -> Type -> Type
+  //.
+  //. Constructor for tuple types of length 2. Arrays are said to represent
+  //. tuples. `['foo', 42]` is a member of `Pair String Number`.
+  var Pair = BinaryTypeWithUrl(
+    'sanctuary-def/Pair',
+    function(x) { return typeEq('Array')(x) && x.length === 2; },
+    function(pair) { return [pair[0]]; },
+    function(pair) { return [pair[1]]; }
+  );
+
+  //# StrMap :: Type -> Type
+  //.
+  //. Constructor for homogeneous Object types.
+  //.
+  //. `{foo: 1, bar: 2, baz: 3}`, for example, is a member of `StrMap Number`;
+  //. `{foo: 1, bar: 2, baz: 'XXX'}` is not.
+  var StrMap = UnaryTypeWithUrl(
+    'sanctuary-def/StrMap',
+    Object_.test.bind(Object_),
+    function(strMap) {
+      return Z.reduce(function(xs, x) { return xs.concat([x]); }, [], strMap);
+    }
   );
 
   //# env :: Array Type
@@ -2576,7 +2572,6 @@
     NonZeroFiniteNumber: NonZeroFiniteNumber,
     NonZeroInteger: NonZeroInteger,
     NonZeroValidNumber: NonZeroValidNumber,
-    None: None,
     Null: Null,
     Nullable: fromUncheckedUnaryType(Nullable),
     Number: Number_,
--- a/index.js
+++ b/index.js
@@ -1788,10 +1788,12 @@
   //. ```
   //.
   //. The `p2` value is a member of `Point3D` as well as `Point`, making it a
-  //. valid argument to `dist`. By default record types allow additional
-  //. properties (which is what allowed `p2` to be a member of `Point`). If one
-  //. wishes for a record type to allow for no additional properties, it can be
-  //. refined using the [`Strict`][] constructor:
+  //. valid argument to `dist`. By default, record types permit the presence
+  //. of additional fields. As a result, p2 is a member of Point as well as
+  //. Point3D.
+  //.
+  //. One could use the [`Strict`][] constructor to define record types which
+  //. do not permit additional fields:
   //.
   //. ```javascript
   //. //    StrictPoint3D :: Type
--- a/index.js
+++ b/index.js
@@ -834,7 +834,9 @@
   //. comprising every object with exactly one field, `name`, of type `String`.
   function Strict(t) {
     function test(x) {
-      return Object.keys(x).length === t.keys.length;
+      var len = 0, o = Object(x);
+      for (var k in o) len += 1; // eslint-disable-line no-unused-vars
+      return t.keys.length === len;
     }
 
     return NullaryType('sanctuary-def/Strict',
--- a/index.js
+++ b/index.js
@@ -353,6 +353,7 @@
     name,       // :: String
     url,        // :: String
     format,     // :: (String -> String, String -> String -> String) -> String
+    parent,     // :: Type
     test,       // :: Any -> Boolean
     keys,       // :: Array String
     types       // :: StrMap { extractor :: a -> Array b, type :: Type }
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -424,9 +429,9 @@
            '#' + stripNamespace(name);
   }
 
-  //  NullaryTypeWithUrl :: (String, Any -> Boolean) -> Type
-  function NullaryTypeWithUrl(name, test) {
-    return NullaryType(name, functionUrl(name), test);
+  //  NullaryTypeWithUrl :: (Type, String, Any -> Boolean) -> Type
+  function NullaryTypeWithUrl(parent, name, test) {
```
davidchambers: I'd prefer `(name, parent, test)` for two reasons:

  - it makes the two constituents of the type's `test` method, `parent` and `test`, adjacent; and
  - having `name` first makes the call sites read better, in my view.

index.js
```diff
@@ -456,17 +461,38 @@
...
+    url: functionUrl('sanctuary-def/Any'),
+    constructor: _Type,
+    test: K(true),
+    validate: Right,
+    format: function(outer) { return outer('Any'); },
+    keys: [],
+    types: {},
+    toString: K('Any')
+  };
+  Any.parent = Any;
```
davidchambers: We could still leverage `NullaryTypeWithUrl` here, I believe:

```javascript
var Any = NullaryTypeWithUrl(null, 'sanctuary-def/Any', null);
Any.parent = Any;
Any.test = K(true);
Any.validate = Right;
```

index.js
```diff
@@ -456,17 +461,38 @@
...
+
+  //# None :: Type
+  //.
+  //. Type with no members.
+  var None = NullaryTypeWithUrl(Any, 'sanctuary-def/None', K(false));
```
davidchambers: Shall we name this `Void` to match Haskell? `None` could be confusing due to its use as a data constructor (of type `Option a`) in Scala.

index.js
```diff
@@ -476,26 +502,41 @@
...
 
   //# FiniteNumber :: Type
   //.
   //. Type comprising every [`ValidNumber`][] value except `Infinity` and
   //. `-Infinity`.
   var FiniteNumber = NullaryTypeWithUrl(
+    ValidNumber,
     'sanctuary-def/FiniteNumber',
-    function(x) { return ValidNumber._test(x) && isFinite(x); }
+    isFinite
```
davidchambers: This change is particularly satisfying. :)

index.js
```diff
@@ -532,28 +571,35 @@
       $types[k] = {extractor: K([]), type: t};
     });
 
-    return new _Type(FUNCTION, '', '', format, test, $keys, $types);
+    return new _Type(FUNCTION, '', '', format,
+                     AnyFunction, K(true), $keys, $types);
```
davidchambers: If it's not possible to have all the arguments on one line, please start each one at the same column. ;)

index.js
Avaq: Agreed :)

index.js
```diff
@@ -476,26 +502,41 @@
...
+  var Number_ = NullaryTypeWithUrl(Any, 'Number', typeofEq('number'));
+
+  //# ValidNumber :: Type
+  //.
+  //. Type comprising every [`Number`][] value except `NaN`.
+  var ValidNumber = NullaryTypeWithUrl(
+    Number_,
+    'sanctuary-def/ValidNumber',
+    function(x) { return !isNaN(x); }
+  );
```
davidchambers: Since it's no longer possible to list the types in alphabetical order, let's decide upon some other sorting rule.

index.js
Avaq: I'm not sure if we need this. I added it to use in INCONSISTENT only, but we could as well use `Any` in there since the `_test` is `K(false)`.

index.js
```diff
@@ -741,28 +787,30 @@
...
 
   //# TypeClass :: Type
   //.
   //. Type comprising every [`TypeClass`][] value.
-  var TypeClass =
-  NullaryTypeWithUrl('TypeClass', typeEq('sanctuary-type-classes/TypeClass'));
+  var TypeClass = NullaryTypeWithUrl(
+    Any,
+    'TypeClass',
+    typeEq('sanctuary-type-classes/TypeClass'));
```
davidchambers: Please wrap the closing `);`.

index.js
Avaq: But my line count! :(

index.js
```diff
@@ -1267,20 +1309,9 @@
...
-  //.   'my-package/NonNegativeInteger',
-  //.   'http://example.com/my-package#NonNegativeInteger',
-  //.   x => $.test([], $.Integer, x) && x >= 0
-  //. );
-  //. ```
-  //.
-  //. Using types as predicates is useful in other contexts too. One could,
-  //. for example, define a [record type][] for each endpoint of a REST API
-  //. and validate the bodies of incoming POST requests against these types.
+  //. Using types as predicates is can be useful. One could, for example,
```
davidchambers: `s/ is//`

I think *powerful* better captures my thought. How about the following?

_Using types as predicates is powerful._

index.js
```diff
@@ -1290,12 +1321,14 @@
   //.
   //. sanctuary-def provides several functions for defining types.
 
-  //# NullaryType :: String -> String -> (Any -> Boolean) -> Type
+  //# NullaryType :: Type -> String -> String -> (Any -> Boolean) -> Type
```
davidchambers: Thank you for remembering to update the documentation! :clap:

Avaq: Not the second time ;)

index.js
davidchambers: Ah, I see! I like the idea of using `Any`. :)

Avaq:
> let's decide upon some other sorting rule

We could sort types by their level in the type tree, alphabetically. So:

1. List of types that have `Any` as a parent, in alphabetical order.
2. List of types that have any of the types from `1`, in alphabetical order.
3. List of types that have any of the types from `2`, in alphabetical order.
4. ...

index.js
davidchambers: This formatting also satisfies my rule:

```javascript
return new _Type(
  FUNCTION, '', '', format, AnyFunction, K(true), $keys, $types
);
```

You're welcome to use this. :)

Avaq:
I prefer grouping "related" types, so perhaps grouping by depth in the type tree is better, we would sort types alphabetically, but ensure that a types children are always listed directly below their parents, that way Number types will be grouped together. It's a slightly more complex rule, but I think will lead to a more readable documentation.

davidchambers:
I like the idea of grouping related types in the readme. It would be nice to have:

  - `Number`
  - `ValidNumber`
  - `NonZeroValidNumber`
  - `FiniteNumber`
  - `NonZeroFiniteNumber`
  - …

I'm not sure how to codify the internal rule which makes this ordering feel natural to me.

Avaq:
What do you think of this new order:

```text
Any
  AnyFunction
  Arguments
  Array
  Boolean
  Date
    ValidDate
  Error
  Inconsistent
  Null
  Number
    ValidNumber
      FiniteNumber
        NegativeFiniteNumber
        NonZeroFiniteNumber
        PositiveFiniteNumber
      Integer
        NegativeInteger
        NonZeroInteger
        PositiveInteger
      NegativeNumber
      NonZeroValidNumber
      PositiveNumber
  Object
  RegexFlags
  RegExp
    GlobalRegExp
  String
  Symbol
  Type
  TypeClass
  Undefined
  Unknown
Function
NonEmpty
Nullable
Pair
StrMap
```

index.js
```diff
@@ -780,22 +726,112 @@
...
+    }
+  );
+
+  //# Pair :: Type -> Type -> Type
+  //.
+  //. Constructor for tuple types of length 2. Arrays are said to represent
+  //. tuples. `['foo', 42]` is a member of `Pair String Number`.
+  var Pair = BinaryTypeWithUrl(
+    'sanctuary-def/Pair',
+    Array_(Any),
```
Avaq: Not sure whether to use `Array Any` or `Array Unknown` here. I think using `Any` is fine.

Avaq:
I think we should hold off implementing this for record types, I think it would be better to leave record types refine `Object`. I think "extending" record types the way you suggested in #128 should be done in a different way. It could have the same API, but not be implemented on top of refinements.

With that out of the way, I think this PR is now "ready".

davidchambers:
> I think "extending" record types the way you suggested in #128 should be done in a different way.

Why do you think so?

index.js
```diff
@@ -398,10 +403,6 @@
   var UNKNOWN       = 'UNKNOWN';
   var VARIABLE      = 'VARIABLE';
 
-  //  Inconsistent :: Type
-  var Inconsistent =
-  new _Type(INCONSISTENT, '', '', always2('???'), K(false), [], {});
-
```
davidchambers: What's the thinking behind moving this definition? Did it feel inconsistent having it apart from the others?

index.js
```diff
@@ -436,15 +437,15 @@
 
   //  UnaryTypeWithUrl ::
   //    (String, Any -> Boolean, t a -> Array a) -> (Type -> Type)
```
davidchambers: This type signature needs to be updated. Let's switch to the other format:

```javascript
  // UnaryTypeWithUrl :: ... -> (Type -> Type)
  function UnaryTypeWithUrl(
    name,           // :: String
    parent,         // :: Type
    test,           // :: Any -> Boolean
    _1              // :: t a -> Array a
  ) {
    ...
  }
```

index.js
```diff
@@ -436,15 +437,15 @@
...
-  function UnaryTypeWithUrl(name, test, _1) {
-    return UnaryType(name, functionUrl(name), test, _1);
+  function UnaryTypeWithUrl(name, parent, test, _1) {
+    return UnaryType(name, functionUrl(name), parent, test, _1);
   }
 
   //  BinaryTypeWithUrl ::
   //    (String, Any -> Boolean, t a b -> Array a, t a b -> Array b) ->
   //      ((Type, Type) -> Type)
```
davidchambers: This type signature needs to be updated too.

index.js
davidchambers: `Array Any` is clearer to me. :thumbsup:

index.js
```diff
@@ -1267,20 +1303,9 @@
...
-  //.   x => $.test([], $.Integer, x) && x >= 0
-  //. );
-  //. ```
-  //.
-  //. Using types as predicates is useful in other contexts too. One could,
-  //. for example, define a [record type][] for each endpoint of a REST API
-  //. and validate the bodies of incoming POST requests against these types.
+  //. Using types as predicates is powerful. One could, for example,
+  //. define a [record type][] for each endpoint of a REST API and
+  //. validate the bodies of incoming POST requests against these types.
```
davidchambers: This wrapping would be more pleasing:

```javascript
  //. Using types as predicates is powerful. One could, for example, define a
  //. [record type][] for each endpoint of a REST API and validate the bodies
  //. of incoming POST requests against these types.
```

index.js
```diff
@@ -1300,6 +1325,8 @@
   //.
   //.   - the documentation URL of `t` (exposed as `t.url`); and
   //.
+  //.   - the parent of `t` (exposed as `t.parent`);
```
davidchambers: Shall we write _the parent <ins>type</ins> of `t`_?

index.js
```diff
@@ -1352,20 +1380,20 @@
...
 
   var CheckedNullaryType =
   def('NullaryType',
       {},
-      [String_, String_, Function_([Any, Boolean_]), Type],
+      [String_, String_, Type, Function_([Any, Boolean_]), Type],
       NullaryType);
 
-  //# UnaryType :: String -> String -> (Any -> Boolean) -> (t a -> Array a) -> (Type -> Type)
+  //# UnaryType :: String -> String -> Type -> (Any -> Boolean) -> ((t a -> Array a) -> (Type -> Type))
```
davidchambers: Why these extra parens?

index.js
```diff
@@ -1573,7 +1610,14 @@
...
                    $2: {extractor: _2, type: $2}};
-      return new _Type(BINARY, name, url, format, test, ['$1', '$2'], types);
+      return new _Type(BINARY,
+                       name,
+                       url,
+                       format,
+                       parent,
+                       test,
+                       ['$1', '$2'],
+                       types);
```
davidchambers: We could remove the `types` variable now that we have more room.

index.js
```diff
@@ -1828,7 +1873,14 @@
...
       var types = {$1: {extractor: K([]), type: $1}};
-      return new _Type(VARIABLE, name, '', format, K(true), ['$1'], types);
+      return new _Type(VARIABLE,
+                       name,
+                       '',
+                       format,
+                       Any,
+                       K(true),
+                       ['$1'],
+                       types);
```
davidchambers: We could remove `types` here too.

test/index.js
```diff
@@ -2475,7 +2479,8 @@ describe('def', function() {
     var Pair = $.BinaryType(
       'my-package/Pair',
       'http://example.com/my-package#Pair',
-      function(x) { return Object.prototype.toString.call(x) === '[object Array]' && x.length === 2; },
+      $.Array($.Any),
+      function(x) { return x.length === 2; },
```
davidchambers: Nice!

index.js
Avaq: It depends on `Any` being defined.

index.js
Avaq: To show that we cannot call the function like `('', '', Any, K(true), x => ...)`

index.js
davidchambers: :thumbsup:

index.js
```diff
@@ -1701,16 +1747,15 @@
     }
 
     function test(x) {
-      return x != null &&
```
Avaq: Indeed. It's similar to how we expect the `length` property to be present in types extending `String`.

test/index.js
```diff
@@ -3133,10 +3139,10 @@ describe('UnaryType', function() {
 
 describe('BinaryType', function() {
 
-  it('is a quinary function', function() {
+  it('is a sexternary function', function() {
```
Avaq: I was looking around for this term for a while. I gave up and figured you'd set it straight ;)

index.js
davidchambers: I'm confused. `UnaryType` takes `(name, url, parent, test, _1)`, doesn't it?

index.js
davidchambers: :thumbsup:

index.js
Avaq:
```
(name, url, parent, test, _1)
                        ^
                        That's the end of the
                        argument list. _1 should
                        be passed into the
                        returned function
```

Avaq:
>> I think "extending" record types the way you suggested in #128 should be done in a different way.

> Why do you think so?

It will never make sense for RecordTypes to refine anything other than Object or RecordType. When I initially thought about it, it seemed like extending a record type is not a form of refinement, but looking at it again, since record types cannot be strict of themselves, I think it actually might be.

Maybe it does make sense to do this, but restrict the "parent" of the RecordType to other RecordTypes (or Object).

index.js
davidchambers: I think you're confusing `_1` (the outer type's extractor) and `$1` (the inner type).

```javascript
//  Maybe :: Type
var Maybe = $.UnaryType(
  'my-package/Maybe',
  'http://example.com/my-package#Maybe',
  $.Any,
  function(x) { return type(x) === 'my-package/Maybe'; },
  function(maybe) { return maybe.isJust ? [maybe.value] : []; }  // _1
);

Maybe($.Integer);  // $.Integer is $1
```

Avaq: I think I am. :man_facepalming:

davidchambers:
> It will never make sense for RecordTypes to refine anything other than Object or RecordType.

I can imagine refining `Array (PossiblyUndefined String)` to represent the record type of which this value is a member:

```javascript
> /abc(def)?/.exec('abc')
["abc", undefined, "index": 0, "input": "abc"]
```

I don't see a reason to treat record types specially.

index.js
Avaq: Furthermore, I was confusing UnaryType for BinaryType when I made this change.

index.js
```diff
@@ -1512,6 +1554,7 @@
   //. const $Pair = $.BinaryType(
   //.   pairTypeIdent,
   //.   'http://example.com/my-package#Pair',
+  //.   Any,
```
Avaq: `$.Any`?

index.js
```diff
@@ -1320,7 +1355,8 @@
   //. const NonZeroInteger = $.NullaryType(
   //.   'my-package/NonZeroInteger',
   //.   'http://example.com/my-package#NonZeroInteger',
-  //.   x => $.test([], Integer, x) && x !== 0
+  //.   Integer,
```
Avaq: `$.Integer`?

index.js
```diff
@@ -1632,26 +1685,31 @@
...
   //.   - an object mapping field name to type.
   //.
   //. For example:
   //.
   //. ```javascript
   //. //    Point :: Type
-  //. const Point = $.RecordType({x: $.FiniteNumber, y: $.FiniteNumber});
+  //. const Point = $.RecordType($.Any, {
+  //.   x: $.FiniteNumber,
+  //.   y: $.FiniteNumber
```
davidchambers: Please add a trailing comma. :)

index.js
```diff
@@ -2485,17 +2550,14 @@
                def);
   }
 
-  var create =
-  def('create',
-      {},
-      [RecordType({checkTypes: Boolean_, env: Array_(Any)}), AnyFunction],
-      _create);
+  var Options = RecordType(Object_, {checkTypes: Boolean_, env: Array_(Any)});
```
davidchambers: Could this be `Array_(Type)` rather than `Array_(Any)`? If there's a good reason for using `Any` rather than `Type`, I've forgotten it. ;)

index.js
davidchambers: Yes, this should be `$.Any`.

index.js
davidchambers: This is referring to the Integer type defined above. One day we should write a new example which does not include a redefinition of a type provided by sanctuary-def.

index.js
Avaq: Oh - I thought we changed that a while ago! When we removed `applyTypes` or something.

index.js
```diff
@@ -1687,7 +1742,58 @@
   //. //
   //. //   The value at position 1 is not a member of ‘{ x :: FiniteNumber, y :: FiniteNumber }’.
   //. ```
-  function RecordType(fields) {
+  //.
+  //. ##### Extending record types
+  //.
+  //. By using the `parent` argument of the RecordType constructor, one can
```
davidchambers: Let's replace `RecordType` with `` `RecordType` `` or `` [`RecordType`][] `` (depending on whether you think it should be a link).

index.js
```diff
@@ -1687,7 +1742,58 @@
...
+  //. define record types which must contain all the same properties of
+  //. another. For example, one could define `Poinst3D` as a refinement of
```
davidchambers: `s/Poinst3D/Point3D/`

index.js
```diff
@@ -1687,7 +1742,58 @@
...
+  //. `Point` from the previous example:
+  //.
+  //. ```js
```
davidchambers: `s/js/javascript/` please, for consistency. :)

index.js
```diff
@@ -1687,7 +1742,58 @@
...
+  //. //    Point3D :: Type
+  //. const Point3D = $.RecordType(Point, {z: $.FiniteNumber});
+  //.
+  //. //    a :: Point
+  //. const a = {x: 0, y: 0};
+  //.
+  //. //    b :: Point, Point3D
+  //. const b = {x: 3, y: 4, z: 5};
```
davidchambers: I think `p1` and `p2` would be slightly better names than `a` and `b`.

index.js
```diff
@@ -1687,7 +1742,58 @@
...
+  //.
+  //. Point.validate(a)   // => Right(a)
+  //. Point3D.validate(a) // => Left({propPath: ['z'], value: undefined})
+  //.
+  //. Point.validate(b)   // => Right(b)
+  //. Point3D.validate(b) // => Right(b)
```
davidchambers: A few minor complaints:

  - I'd like to see semicolons here;
  - I'd like to see (at least) two spaces before the `//` beginning a line comment; and
  - I'd like to see `Right({x: 0, y: 0})`/`Right({x: 3, y: 4, z: 5})` rather than `Right(a)`/`Right(b)`.

index.js
```diff
@@ -1687,7 +1742,58 @@
...
+  //.
+  //. dist(a, b);
+  //. // => 5
+  //. ```
+  //.
+  //. The `b` value is a member of `Point3D` as well as `Point`, allowing one
+  //. to call `dist` on it. By default record types allow additional
```
davidchambers: I suggest replacing _allowing one to call `dist` on it_ with _making it a valid argument to `dist`_. The former sounds like a method invocation.

index.js
```diff
@@ -1687,7 +1742,58 @@
...
+  //. properties, permitting record "extension". If one wishes for a record
```
davidchambers: What does _permitting record "extension"_ mean, exactly? Could this clause be dropped?

index.js
```diff
@@ -1687,7 +1742,58 @@
...
+  //. type to allow for no additional properties, it can be refined as such:
+  //.
+  //. ```js
```
davidchambers: `s/js/javascript/`

index.js
```diff
@@ -1687,7 +1742,58 @@
...
+  //. //    strict :: Type -> Type
+  //. const strict = t => $.NullaryType(
```
davidchambers: Should this be named `Strict`, to match [`$.NonEmpty`][1]? Similarly, could it be a unary type so one would define `Strict Point` rather than `StrictPoint`?


[1]: https://github.com/sanctuary-js/sanctuary-def#NonEmpty

index.js
```diff
@@ -1687,7 +1742,58 @@
...
+  //.   'Strict' + t.name,
+  //.   '',
+  //.   t,
+  //.   x => Object.keys(x).length === t.keys.length
+  //. );
+  //.
+  //. //    StrictPoint3D :: Type
+  //. const StrictPoint3D = strict(Point3D)
```
davidchambers: `s/$/;/`

index.js
```diff
@@ -1687,7 +1742,58 @@
...
+  //.
+  //. //    a :: Point, Point3D
+  //. const a = {x: 1, y: 2, z: 3, color: 'red'}
```
davidchambers: `s/$/;/`

index.js
```diff
@@ -1687,7 +1742,58 @@
...
+  //.
+  //. Point3D.validate(a)       // => Right(a)
+  //. StrictPoint3D.validate(a) // => Left({propPath: [], value: a})
```
davidchambers: My minor complaints from an earlier comment apply here too.

index.js
```diff
@@ -2485,17 +2598,14 @@
                def);
   }
 
-  var create =
-  def('create',
-      {},
-      [RecordType({checkTypes: Boolean_, env: Array_(Any)}), AnyFunction],
-      _create);
+  var Options = RecordType(Object_, {checkTypes: Boolean_, env: Array_(Type)});
+  var create = def('create', {}, [Options, AnyFunction], _create);
```
davidchambers: I'd prefer this arrangement to make it clear that the record type is not used elsewhere:

```javascript
  var create =
  def('create',
      {},
      [RecordType(Object_, {checkTypes: Boolean_, env: Array_(Type)}),
       AnyFunction],
      _create);
```

test/index.js
```diff
@@ -236,13 +240,13 @@ describe('create', function() {
            TypeError,
            'Invalid value\n' +
            '\n' +
-           'create :: { checkTypes :: Boolean, env :: Array Any } -> Function\n' +
-           '          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n' +
+           'create :: { checkTypes :: Boolean, env :: Array Type } -> Function\n' +
+           '          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n' +
```
davidchambers: This is a nice improvement. :)

test/index.js
```diff
@@ -1200,14 +1204,14 @@ describe('def', function() {
...
   it('supports record types', function() {
     eq(typeof $.RecordType, 'function');
-    eq($.RecordType.length, 1);
-    eq($.RecordType.toString(), 'RecordType :: StrMap Type -> Type');
+    eq($.RecordType.length, 2);
+    eq($.RecordType.toString(), 'RecordType :: Type -> StrMap Type -> Type');
 
     //  Point :: Type
-    var Point = $.RecordType({x: $.Number, y: $.Number});
+    var Point = $.RecordType($.Any, {x: $.Number, y: $.Number});
```
davidchambers: Every record type defined in the test suite currently extends `$.Any`. It would be good to include your `Point3D` example here as well.

index.js
```diff
@@ -1687,7 +1742,62 @@
...
+  //. Point.validate(p2);    // => Right({x: 3, y: 4, z: 5})
+  //. Point3D.validate(p2);  // => Right({x: 3, y: 4, z: 5})
+  //.
+  //. dist(p1, p2);
+  //. // => 5
+  //. ```
+  //.
+  //. The `p2` value is a member of `Point3D` as well as `Point`, making it a
+  //. valid argument to `dist`. By default record types allow additional
+  //. properties (which is what allowed `p2` to be a member of `Point`). If one
```
Avaq: I want readers to be aware that permitting additional properties is a prerequisite to allowing *record extension* as described in the prior section. Maybe this more concrete clause is more to your liking?

index.js
```diff
@@ -1687,7 +1742,62 @@
...
+  //. refined as such:
+  //.
+  //. ```javascript
+  //. //    strict :: Type -> Type
+  //. const strict = t => $.NullaryType(
+  //.   'Strict' + t.name,
+  //.   '',
+  //.   t,
+  //.   x => Object.keys(x).length === t.keys.length
+  //. );
```
Avaq: How do you propose this be made into a UnaryType? I see one hacky way:

```js
//    Strict :: Type -> Type
const Strict = t => $.UnaryType(
  'Strict' + t.name,
  t.url,
  t,
  x => Object.keys(x).length === t.keys.length,
  x => [x]
)(t); //<- That's unfortunate
```

The "proper" way leads to a dead end:

```js
//    Strict :: Type -> Type
const Strict = $.UnaryType(
  'my-package/Strict',  //<- That's unfortunate
  'http://example.com/#strict',  //<- That's unfortunate
  $.Any,  //<- That's unfortunate
  x => Object.keys(x).length === `???`,  //<- That's impossible
  x => [x]
);
```

I currently think `strict` makes more sense as a "common refinement" (which justifies its inclusion in the library), and I probably would have also made `nonEmpty` in that way.

Avaq: The advantage of having NonEmpty be a UnaryType is that `"abc"` is automatically a member of it, correct? So we don't have to explicitly create `NonEmptyString` and add it to the environment.

Avaq: Maybe the `test` could be something like:

```js
x => {
  const len = Object.keys(x).length;
  return determineActualTypesStrict(x).every(type => type.keys.length === len);
}
```

Avaq: The first "hacky" approach will not have the same benefits as `NonEmpty`. We have to find an approach where `Strict(Unknown)` has members.

Avaq: Altough, I don't have to add my Strict type to the environment anyway, I can create it at run time without drawback, I think:

```js
strict(User).validate(userInput)
```

davidchambers: How about this approach?

```javascript
//    Strict :: Type -> Type
const Strict = $.UnaryType(
  'my-package/Strict',
  'http://example.com/my-package#Strict',
  $.Any,
  function(x) {
    const t = this.types.$1.type;
    return t.type === 'RECORD' &&
           $.test([], t, x) &&
           Object.keys(x).length === t.keys.length;
  },
  strict => [strict]
);
```

index.js
davidchambers: I :heart: the concrete clause. In fact, I would lead with it:

_By default, record types permit the presence of additional fields. As a result, `p2` is a member of `Point` as well as `Point3D`._

_One could define `Strict :: Type -> Type` and use it to define record types which do not permit additional fields:_

index.js
Avaq: Woah. Using `this` inside a `test` function is new, correct? There are some places where we're passing `t._test` down into a new Type, I'm not sure if the consequences of relying on `this` will include breakage of code relying on the "purity" (*`this`-lessness*?) of `_test`. If it doesn't, then the fact that we can use `this` to refer to the type of a value being `test`ed is actually quite neat.

I do still doubt whether using `UnaryType` for this purpose is useful though. When implementing this as a type refinement, we get the elegance of `strict(User)` being a refinement of `User`. When implementing this as a `UnaryType`, I'm not sure if we gain anything; I believe that adding `Strict(Unknown)` to the environment will not cause `{name: "Bob"} :: User, Strict User`, because `Unknown.type !== 'RECORD'`. This differs from `NonEmpty`, as *its* check does not rely on the type of `types.$1`.

Avaq: I have a much stronger argument:

```js
> var User = $.RecordType($.Any, {name: $.NonEmpty($.String)})

> User.validate({name: ''})
Left({ value: '', propPath: [ 'name' ] })

> strict(User).validate({name: ''});
Left({ value: '', propPath: [ 'name' ] })

> Strict(User)
Left({ value: { name: '' }, propPath: [] })
```

Or in words: Your `UnaryType` relies on `$.test`, and the purpose of parent types was to fix `validate` for those types that relied on `test`-ing another type. Actually, the purpose of this PR was to fix #128, and in particular, to address my comment there: https://github.com/sanctuary-js/sanctuary-def/issues/128#issuecomment-287705497

index.js
```diff
@@ -2538,6 +2668,7 @@
     PositiveNumber: PositiveNumber,
     RegExp: RegExp_,
     RegexFlags: RegexFlags,
+    Strict: Strict,
     StrMap: fromUncheckedUnaryType(StrMap),
     String: String_,
```
davidchambers: I've been using `Array#sort` to determine order:

```javascript
> ['Strict', 'StrMap', 'String'].sort()
['StrMap', 'Strict', 'String']
```

Please also move the definition of `Strict`.

Avaq: I tried, but found that my editors built-in sorting mechanism changed the position of String.

test/index.js
```diff
@@ -1345,6 +1357,22 @@ describe('def', function() {
            'Since there is no type of which all the above values are members, the type-variable constraint has been violated.\n');
   });
 
+  it('support "strict" record types', function() {
```
davidchambers: `s/support/supports/`

test/index.js
```diff
@@ -1345,6 +1357,22 @@ describe('def', function() {
            'Since there is no type of which all the above values are members, the type-variable constraint has been violated.\n');
   });
 
+  it('supports "strict" record types', function() {
+
```
davidchambers: For consistency, please remove this empty line and the one before the closing `});`.

test/index.js
```diff
@@ -1345,6 +1357,22 @@ describe('def', function() {
...
+    var Point = $.RecordType($.Any, {x: $.Number, y: $.Number});
+
+    //  StrictPoint :: Type
+    var StrictPoint = $.Strict(Point);
+
+    //  p :: Point
+    var p = {x: 1, y: 2, z: 3};
+
+    eq($.test([], Point, p), true);
+    eq($.test([], StrictPoint, p), false);
```
davidchambers: I'd like to see four assertions here:

```javascript
    eq($.test([], Point, {x: 1, y: 2}), true);
    eq($.test([], Point, {x: 1, y: 2, z: 3}), true);
    eq($.test([], StrictPoint, {x: 1, y: 2}), true);
    eq($.test([], StrictPoint, {x: 1, y: 2, z: 3}), false);
```

index.js
```diff
@@ -780,22 +734,129 @@
...
+    function(pair) { return [pair[0]]; },
+    function(pair) { return [pair[1]]; }
+  );
+
+  //# Strict :: Type -> Type
+  //.
+  //. Constructor for strict record types. For example:
+  //. `$.Strict($.Record($.Any, {name: $.String}))`, is the type comprising
+  //. every object with a `name :: String` property, without any other
+  //. properties.
```
davidchambers: Do you like this variant?

_Constructor for strict record types._

_`$.Strict($.Record($.Any, {name: $.String}))`, for example, is the type comprising every object with exactly one field, `name`, of type `String`._

index.js
```diff
@@ -780,22 +734,129 @@
...
+  function Strict(t) {
+    function test(x) {
+      return Object.keys(x).length === t.keys.length;
```
davidchambers: `Object.keys` ignores inherited properties. Is this desirable?

Avaq: I thought about that, and I don't think we can make a sensible Strict constructor otherwise.

index.js
```diff
@@ -1687,7 +1759,59 @@
...
+  //. Point3D.validate(p2);  // => Right({x: 3, y: 4, z: 5})
+  //.
+  //. dist(p1, p2);
+  //. // => 5
+  //. ```
+  //.
+  //. The `p2` value is a member of `Point3D` as well as `Point`, making it a
+  //. valid argument to `dist`. By default, record types permit the presence
+  //. of additional fields. As a result, p2 is a member of Point as well as
+  //. Point3D.
```
davidchambers: `p2`, `Point`, and `Point3D` should be wrapped in backticks.

Avaq: Darned copy-paste ;)

index.js
davidchambers: Is using `for...in` impractical for some reason?

```javascript
function O() {}
var o = new O();
o.direct = true;
O.prototype.indirect = true;
Object.keys(o);
// => ['direct']
(function() { var keys = []; for (var k in o) keys.push(k); return keys; }());
// => ['direct', 'indirect']
```

Avaq: We use `keys.every(k => k in o)` as a `test` for record types. Will this lead to problems?

```js
> 'then' in new Promise(() => {})
true

> (function(o) { var keys = []; for (var k in o) keys.push(k); return keys; }(new Promise(() => {})))
[]
```

This would mean that:

```js
const Thenable = $.RecordType({then: $.AnyFunction});
const StrictThenable = $.Strict(Thenable);

Thenable.validate(new Promise(noop)) // => Right
StrictThenable.validate(new Promise(noop)) // => Left
```

I'm not sure if that's desired. Using enumerable keys over "own" keys is an improvement though, so I think I should implement your suggestion nonetheless.

index.js
```diff
@@ -797,17 +797,16 @@
...
+    function test(x) {
       return Z.Monoid.test(x) &&
              Z.Setoid.test(x) &&
              !Z.equals(x, Z.empty(x.constructor));
-    },
-    function(monoid) { return [monoid]; }
-  );
+    }
+    function extract(monoid) { return [monoid]; }
+    var t = UnaryTypeWithUrl('sanctuary-def/NonEmpty', parent, test, extract);
```
Avaq: Using `parent` as an *actual* parent type, like I'm doing here, seems like a good idea. Now `NonEmpty String` will (correctly) be a "refinement" of `String`.

However, as indicated by the failing unit test, it causes a value of `[1]` for `$NonEmpty($Array($String))`, for example, to generate this error message:

```text
NonEmpty (Array String)
         ^^^^^^^^^^^^^^
               1

1) 1 :: Number

The value at position 1 is not a member of ‘Array String’. 
```

Where we would want:

```text
NonEmpty (Array String)
                ^^^^^^
                  1

1) 1 :: Number

The value at position 1 is not a member of ‘String’.
```

I'm not sure why this happens. Replacing `parent` with `Any` solves the issue.

davidchambers: The problem is that

```javascript
$.NonEmpty($.Array($.String)).validate([1, 2, 3]).value.propPath
```

evaluates to `['$1']` rather than `['$1', '$1']` because we first validate the parent type and, if validation fails, make no adjustment to `propPath` to account for the extra level of nesting introduced by `$.NonEmpty`.

Avaq: That seems like an issue that will bite us in some other way, one day. I don't yet see a solution.

Avaq: In it's general form, I think the issue could be described as: *When the validation of a refined type fails on its parent type, the reported `propPath` may not correctly correspond to the location of the value as it exists in the refined type*

I think we might even encounter this issue if we refine a record type to be "strict", as has always been the suggestion:

```js
const User = $.RecordType({ name: $.String, password: $.String });
const StrictUser = $.NullaryType('', '', User, x => Object.keys(x).length === User.keys.length);

const validation = StrictUser.validate({ name: 'bob', password: 1 });
// propPath = [ 'password' ]

StrictUser.types[propPath[0]] // => undefined
// the above expression should never return undefined, as the propPath must
// correspond to the types mapping. It does because our propPath is actually
// corresponding to another types mapping!
```

Avaq: The same problem, biting us somewhere else:

```js
const Point = $.RecordType($.Any, { x: $.Number, y: $.Number });
const Point3D = $.RecordType(Point, { z: $.Number });

Point3D.validate({ x: 1, y: 'two', z: 3 });
// propPath = [ 'y' ]

Point3D.types[propPath[0]] // => undefined
```

Avaq: This is particularly problematic, because the property names described by `propPath` might exist in both the parent type and the refinement, and `validate` does not tell us which it was.

davidchambers:
Do you remember where we left this, Aldwin? In light of sanctuary-js/sanctuary#464 I'm very excited by the possibility of merging this pull request. :)

Avaq:
Hi David. I don't think this will be ready to merge anytime soon. We left it at the point where we realised that this brings more problems than it solves. Here's some exempts from our Gitter conversation at the time:

> Hi David! I'm having trouble again with the `Strict`/`NonEmpty` case. It seemed to have a clear solution last time:
>
> We created type refinements with the purpose of defining strict record types in terms of them, therefore it goes without saying that `Strict` should be a helper function that takes away the boilerplate of doing so.
>
> We then moved on to conclude that `NonEmpty` is very similar to `Strict`, and should probably follow the same logic. However, after defining `NonEmpty` this way, I've run into an issue that could not be encountered with `Strict`, to do with `t.name`. Look at this definition:
>
> ```js
>  function NonEmpty(t) {
>    return NullaryType('sanctuary-def/NonEmpty',
>                       functionUrl('NonEmpty'),
>                       t,
>                       function test(x) {
>                         return Z.Monoid.test(x) &&
>                                Z.Setoid.test(x) &&
>                                !Z.equals(x, Z.empty(x.constructor));
>                       });
>  }
> ```
>
> This will cause the following:
>
> ```js
> const head = def('head', {}, [$NonEmpty($Array($a)), $a], xs => xs[0]);
> head.toString()
> ```
>
> ```diff
> -head :: NonEmpty (Array a) -> a
> +head :: NonEmpty -> a
> ```
>
> We have not run into this issue with `Strict`, simply because we don't render the name of record types, and have not tested against it. However, I suspect that after making a record type "Strict", it will no longer be rendered very informatively in error messages. All of this leads me to conclude that:
>
> 1. Using a `UnaryType` for `NonEmpty` makes sense, because we keep all type information for the inner type.
> 2. Using a `NullaryType` refinement for `Strict` is not good enough. Especially when we ever add names to record types.
>
> I'm going to explore using a `UnaryType` for both `NonEmpty` and `Strict`, and see if I can still get them to be a refinement as well as a "container" of the input type.

----

> I've modified `Type#validate` to include a `expType :: Type` field in the returned record, for example in `$.Integer.validate('hello')`, this would have a value of `$.Number`, because that's the actual type for which our validation failed. The `propPath` then always applies to the `expType`, and we can even deduce that:
> 
> ```text
> 1) "hello" :: String
> 
> The value at position 1 is not a member of Integer, because it is not a Number.
> ```
>
> ...for example. But there are a lot of changes to be made to the code which builds on `Type#validate` before we are at that point.
>
>> I hope you're able to find a way to make $.Strict a unary type.

----

> Type refinements worked so well on the surface, but upon taking a closer look I find that many things are broken under the current approach. Take this for example:
> 
> ```js
> const Point = $.RecordType($.Any, {x: $.Number, y: $.Number});
> const Point3D = $.RecordType(Point, {z: $.Number});
> 
> const StrictPoint3D = $.Strict(Point3D);
> ```
>
> We haven't considered that `StrictPoint3D` will have 0 members. The `Strict` wrapper expects an object with one key, because `Point3D.keys.length === 1`, and when an object with one key is passed, it's invalidated by the parent, which expects two keys.
>
> It's a similar problem to what we faced before. The meta-data on `Point3D` does not reflect the "full" meta-data of the combined `Point`+`Point3D`.
>
>> It seems we need to walk the tree to construct the union ourselves. We'll also need to decide what to do if we encounter a contradiction (e.g. `{foo: $.Number}` and `{foo: $.String}`).

----

That's the last few thoughts I had on the subject, and where things came to an unfortunate halt. 

Incidentally, our Gitter conversation is a treasure trove of ideas! Here's some:

1. I called https://github.com/sanctuary-js/sanctuary/pull/464 in July!
    > I've been thinking about the purpose of `env`, and there is one burning question that arises: Why does sanctuary add types like `$.FiniteNumber` to the environment? [...] We're adding types to the environment for the sole purpose of showing nicer error messages. For the purpose of *type variables*, the only types that need to be in the environment are direct descendants of `Any`. [This] means we can optimize the amount of checks we're doing by ignoring any types where `t.parent !== Any`.
    >
    > And when it comes to *reporting* all known types of a value, why do we limit ourselves the the environment? It starts to become attractive to simply add *all* types to the environment. The more types Sanctuary knows about, the more precisely an error message will report what the types of a value are at no extra run-time cost[, because of the previous idea].

2. Or a change to the `toMaybe` API to work more like `encase`. You asked me to open an issue and I forgot, so here it is: https://github.com/sanctuary-js/sanctuary/issues/465

davidchambers:
Shall we close this, Aldwin?

Avaq:
Let's do it. The effort is not lost, because it has gotten me familiar with the sanctuary-def code. :)

PR #163, opened 2017-06-27
replace Unknowns with free type variables in string representations

davidchambers:
Commit message:

> Currently `???` is used for both Inconsistent and Unknown:
>
>     ['foo', true, 42] :: Array ???
>     []                :: Array ???
>
> This commit visually differentiates the two types:
>
>     ['foo', true, 42] :: Array ???
>     []                :: Array a
>
> The type variable is chosen to avoid collisions with variables in the corresponding type signature.

This pull request was motivated by comments in #160.

### DIFF
--- a/index.js
+++ b/index.js
@@ -787,7 +787,8 @@
   //.   - `List (List (List Number))`
   //.   - `List (List (List String))`
   //.   - `...`
-  var Unknown = new _Type(UNKNOWN, '', '', always2('???'), K(true), [], {});
+  var Unknown =
+  new _Type(UNKNOWN, '', '', always2('Unknown'), K(true), [], {});
 
   //# ValidDate :: Type
   //.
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -2095,11 +2096,38 @@
...
+    for (var code = a; code <= z; code += 1) {
+      chars.push(String.fromCharCode(code));
+    }
+    return chars;
+  }());
+
+  //  showType :: TypeInfo -> Type -> String
+  function showType(typeInfo) {
+    return function(t) {
+      var names = Z.chain(typeVarNames, typeInfo.types);
```
Avaq: You can hoist this computation so that it will not be executed for every type.

index.js
```diff
@@ -2095,11 +2096,38 @@
...
+      var $free = typeVarChars.filter(function(name) {
+        return names.indexOf(name) < 0;
+      });
+      return unless(
+        t.type === FUNCTION || t.type === RECORD || isEmpty(t.keys),
+        stripOutermostParens,
+        String(t).replace(/\bUnknown\b/g, function() { return $free.shift(); })
```
Avaq: If you use `$free[idx++]` instead, you can also hoist the computation for `$free`.

index.js
```diff
@@ -2109,17 +2137,24 @@
...
+    typeInfo,       // :: TypeInfo
+    values,         // :: Array Any
+    pos             // :: Integer
+  ) {
     return String(pos) + ')  ' + Z.map(function(x) {
       var types = determineActualTypesLoose(env, [x]);
-      return Z.toString(x) + ' :: ' + Z.map(showType, types).join(', ');
+      return Z.toString(x) +
+             ' :: ' +
+             Z.map(showType(typeInfo), types).join(', ');
```
Avaq: You could assign the result of `showType(typeInfo)` to a variable to make better use of the optimizations proposed before.

davidchambers:
I've made some changes based on your feedback, Aldwin. Please have another look. :)

index.js
```diff
@@ -2095,11 +2096,29 @@
...
+    var names = Z.chain(typeVarNames, typeInfo.types);
+    return function(t) {
+      var code = 'a'.charCodeAt(0);
+      return unless(
+        t.type === FUNCTION || t.type === RECORD || isEmpty(t.keys),
+        stripOutermostParens,
+        String(t).replace(/\bUnknown\b/g, function() {
+          do var name = String.fromCharCode(code++);
+          while (names.indexOf(name) >= 0);
+          return name;
```
davidchambers: The rarely seen `do...while`!

PR #179, opened 2017-12-17
abandon Ramda-style currying in favour of simple currying

davidchambers:
Commit message:

> `def` has served two roles:
>
>   - currying an uncurried implementation function; and
>   - performing type checking at run time.
>
> At the time of this project's inception in 2015 it seemed unreasonable to require users to provide curried implementation functions. Defining curried functions manually was onerous prior to ES6:
>
> ```javascript
> function(x) {
>   return function(y) {
>     return function(z) {
>       return ...;
>     };
>   };
> }
> ```
>
> With ES6, defining curried functions is trivial:
>
> ```javascript
> x => y => z => ...
> ```
>
> The landscape has changed since 2015, and it's now reasonable to assume that users are targeting ES6. Furthermore, the advantages of Ramda-style currying over simple currying do not justify the additional complexity. There's no compelling reason for `def` to continue to serve two roles.
>
> User-facing changes:
>
>   - `$` functions must now be applied to arguments one at a time;
>   - `def` now requires a curried implementation function;
>   - `def` no longer imposes an arbitrary arity limit;
>   - `def` functions must now be applied to arguments one at a time; and
>   - `__` has been removed (simple currying precludes placeholders).
>
> Internal changes:
>
>   - the `checkTypes` option is now checked in just one place; and
>   - `def` is now essentially a no-op if type checking is disabled.

This pull request embraces the `f (x) (y)` style proposed in sanctuary-js/sanctuary#438.

The changes to the test suite will be easier to review once #178 has been merged.

### DIFF
--- a/index.js
+++ b/index.js
@@ -18,7 +18,7 @@
 //. It is conventional to import the package as `$`:
 //.
 //. ```javascript
-//. const $ = require('sanctuary-def');
+//. const $ = require ('sanctuary-def');
 //. ```
 //.
 //. The next step is to define an environment. An environment is an array
@@ -28,13 +28,13 @@
 //.
 //. ```javascript
 //. //    Integer :: Type
-//. const Integer = ...;
+//. const Integer = '...';
 //.
 //. //    NonZeroInteger :: Type
-//. const NonZeroInteger = ...;
+//. const NonZeroInteger = '...';
 //.
 //. //    env :: Array Type
-//. const env = $.env.concat([Integer, NonZeroInteger]);
+//. const env = $.env.concat ([Integer, NonZeroInteger]);
 //. ```
 //.
 //. Type constructors such as `List :: Type -> Type` cannot be included in
@@ -43,13 +43,13 @@
 //.
 //. ```javascript
 //. //    env :: Array Type
-//. const env = $.env.concat([
-//.   List($.Number),               // :: Type
-//.   List($.String),               // :: Type
-//.   List(List($.Number)),         // :: Type
-//.   List(List($.String)),         // :: Type
-//.   List(List(List($.Number))),   // :: Type
-//.   List(List(List($.String))),   // :: Type
+//. const env = $.env.concat ([
+//.   List ($.Number),                // :: Type
+//.   List ($.String),                // :: Type
+//.   List (List ($.Number)),         // :: Type
+//.   List (List ($.String)),         // :: Type
+//.   List (List (List ($.Number))),  // :: Type
+//.   List (List (List ($.String))),  // :: Type
 //. ]);
 //. ```
 //.
@@ -58,13 +58,13 @@
 //.
 //. ```javascript
 //. //    env :: Array Type
-//. const env = $.env.concat([List($.Unknown)]);
+//. const env = $.env.concat ([List ($.Unknown)]);
 //. ```
 //.
 //. The next step is to define a `def` function for the environment:
 //.
 //. ```javascript
-//. const def = $.create({checkTypes: true, env: env});
+//. const def = $.create ({checkTypes: true, env});
 //. ```
 //.
 //. The `checkTypes` option determines whether type checking is enabled.
@@ -72,9 +72,9 @@
 //. during development. For example:
 //.
 //. ```javascript
-//. const def = $.create({
+//. const def = $.create ({
 //.   checkTypes: process.env.NODE_ENV === 'development',
-//.   env: env,
+//.   env,
 //. });
 //. ```
 //.
@@ -83,37 +83,49 @@
 //. ```javascript
 //. //    add :: Number -> Number -> Number
 //. const add =
-//. def('add', {}, [$.Number, $.Number, $.Number], (x, y) => x + y);
+//. def ('add')
+//.     ({})
+//.     ([$.Number, $.Number, $.Number])
+//.     (x => y => x + y);
 //. ```
 //.
 //. `[$.Number, $.Number, $.Number]` specifies that `add` takes two arguments
-//. of type `Number` and returns a value of type `Number`.
+//. of type `Number`, one at a time, and returns a value of type `Number`.
 //.
-//. Applying `add` to two arguments gives the expected result:
+//. Applying `add` to two arguments, one at a time, gives the expected result:
 //.
 //. ```javascript
-//. add(2, 2);
+//. add (2) (2);
 //. // => 4
 //. ```
 //.
-//. Applying `add` to greater than two arguments results in an exception being
+//. Applying `add` to multiple arguments at once results in an exception being
 //. thrown:
 //.
 //. ```javascript
-//. add(2, 2, 2);
-//. // ! TypeError: ‘add’ requires two arguments; received three arguments
+//. add (2, 2, 2);
+//. // ! TypeError: ‘add’ applied to the wrong number of arguments
+//. //
+//. //   add :: Number -> Number -> Number
+//. //          ^^^^^^
+//. //            1
+//. //
+//. //   Expected one argument but received three arguments:
+//. //
+//. //     - 2
+//. //     - 2
+//. //     - 2
 //. ```
 //.
-//. Applying `add` to fewer than two arguments results in a function
-//. awaiting the remaining arguments. This is known as partial application.
-//. Partial application is convenient as it allows more specific functions
-//. to be defined in terms of more general ones:
+//. Applying `add` to one argument produces a function awaiting the remaining
+//. argument. This is known as partial application. Partial application allows
+//. more specific functions to be defined in terms of more general ones:
 //.
 //. ```javascript
 //. //    inc :: Number -> Number
-//. const inc = add(1);
+//. const inc = add (1);
 //.
-//. inc(7);
+//. inc (7);
 //. // => 8
 //. ```
 //.
@@ -121,16 +133,15 @@
 //. errors. Consider the following function:
 //.
 //. ```javascript
-//. //    _add :: (Number, Number) -> Number
-//. const _add = (x, y) => x + y;
+//. //    _add :: Number -> Number -> Number
+//. const _add = x => y => x + y;
 //. ```
 //.
-//. The type signature indicates that `_add` takes two arguments of type
-//. `Number`, but this is not enforced. This allows type errors to be silently
-//. ignored:
+//. The type signature indicates that `_add` takes arguments of type `Number`,
+//. but this is not enforced. This allows type errors to be silently ignored:
 //.
 //. ```javascript
-//. _add('2', '2');
+//. _add ('2') ('2');
 //. // => '22'
 //. ```
 //.
@@ -138,7 +149,7 @@
 //. types:
 //.
 //. ```javascript
-//. add('2', '2');
+//. add ('2') ('2');
 //. // ! TypeError: Invalid value
 //. //
 //. //   add :: Number -> Number -> Number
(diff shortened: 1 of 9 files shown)


index.js
```diff
@@ -1365,20 +1316,24 @@
...
   //. ```
-  function NullaryType(name, url, test) {
-    function format(outer, inner) {
-      return outer(stripNamespace(name));
-    }
-    return new _Type(NULLARY, name, url, format, test, [], {});
+  function NullaryType(name) {
+    return function(url) {
+      return function(test) {
+        function format(outer, inner) {
```
Avaq: It looks like the `format` function could be shared between different applications of `NullaryType (name)`.

index.js
```diff
@@ -1456,38 +1413,47 @@
...
-               inner('$1')(String($1)) + outer(')');
-      }
-      var types = {$1: {extractor: _1, type: $1}};
-      return new _Type(UNARY, name, url, format, test, ['$1'], types);
+  function UnaryType(name) {
+    return function(url) {
+      return function(test) {
+        return function(_1) {
+          return function($1) {
+            function format(outer, inner) {
```
Avaq: ditto

index.js
```diff
@@ -1577,46 +1549,67 @@
...
-                   $2: {extractor: _2, type: $2}};
-      return new _Type(BINARY, name, url, format, test, ['$1', '$2'], types);
+  function BinaryType(name) {
+    return function(url) {
+      return function(test) {
+        return function(_1) {
+          return function(_2) {
+            return function($1) {
+              return function($2) {
+                function format(outer, inner) {
```
Avaq: ditto

index.js
```diff
@@ -1868,45 +1870,50 @@
...
-        return outer('(' + name + ' ') + inner('$1')(String($1)) + outer(' ') +
-                                         inner('$2')(String($2)) + outer(')');
-      }
-      var keys = ['$1', '$2'];
-      var types = {$1: {extractor: K([]), type: $1},
-                   $2: {extractor: K([]), type: $2}};
-      return new _Type(VARIABLE, name, '', format, K(true), keys, types);
+    return function($1) {
+      return function($2) {
+        function format(outer, inner) {
```
Avaq: ditto

index.js
```diff
@@ -2429,147 +2368,195 @@
...
   function fromUncheckedUnaryType(typeConstructor) {
     var t = typeConstructor(Unknown);
     var _1 = t.types.$1.extractor;
-    return CheckedUnaryType(t.name, t.url, t._test, _1);
+    return CheckedUnaryType(t.name)(t.url)(t._test)(_1);
   }
 
-  //  fromUncheckedBinaryType :: ((Type, Type) -> Type) ->
+  //  fromUncheckedBinaryType :: (Type -> Type -> Type) ->
   //                             (Type -> Type -> Type)
```
Avaq: Brackets for clarity?

test/index.js
```diff
@@ -1,81 +1,97 @@
...
+    (x => y => [x, y]);
+
+//    $3 :: a -> a -> a -> Array a
+const $3 =
+def ('$3')
+    ({})
+    ([a, a, a, $.Array (a)])
+    (x => y => z => [x, y, z]);
+
+//    $26 :: a -> a -> a -> a -> a -> a -> a -> a -> a -> a -> a -> a -> a -> a -> a -> a -> a -> a -> a -> a -> a -> a -> a -> a -> a -> a -> Array a
```
Avaq: Haha

test/index.js
```diff
@@ -1,81 +1,97 @@
...
-function notImplemented() { throw new Error('Not implemented'); }
-
-//  throws :: (Function, TypeRep a, String) -> Undefined
-function throws(f, type, message) {
-  assert.throws(f, function(err) {
-    return err.constructor === type && err.message === message;
-  });
+//    eq :: (a, b) -> Undefined !
+function eq(actual, expected) {
+  assert.strictEqual (arguments.length, eq.length);
```
Avaq: Curious that you use the new style in the examples and the tests, but not the actual source.

index.js
```diff
@@ -58,23 +58,23 @@
 //.
 //. ```javascript
 //. //    env :: Array Type
-//. const env = $.env.concat([List($.Unknown)]);
+//. const env = $.env .concat ([List ($.Unknown)]);
```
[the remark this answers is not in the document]
davidchambers: The space after the method name is to appease the linter. If we want to write `f (x)` the linter wants us to write `foo.bar (baz)` as well. I was experimenting with surrounding method names with spaces to make them appear more like infix operators: `xs .concat (ys)`. What do you think?

index.js
```diff
@@ -402,6 +393,10 @@
   var Inconsistent =
   new _Type(INCONSISTENT, '', '', always2('???'), K(false), [], {});
 
+  //  NoArguments :: Type
+  var NoArguments =
+  new _Type(NO_ARGUMENTS, '', '', always2('()'), K(true), [], {});
```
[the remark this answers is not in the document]
davidchambers: `NoArguments` and `Unit` represent related but distinct ideas.

Were we to define `const Unit = {};` and the accompanying Type value, `f ()` and `f (Unit)` would still be different expressions: one applies `f` to no arguments; the other applies `f` to one argument.

It could be confusing to name the above Type value `Unit` as we may one day wish to define a Type value equivalent to Haskell's `()`. This would logically be named `Unit`.

index.js
```diff
@@ -498,16 +493,23 @@
...
   //. Constructor for Function types.
   //.
   //. Examples:
   //.
-  //.   - `$.Function([$.Date, $.String])` represents the `Date -> String`
+  //.   - `$.Function ([$.Date, $.String])` represents the `Date -> String`
   //.     type; and
-  //.   - `$.Function([a, b, a])` represents the `(a, b) -> a` type.
-  function Function_(types) {
+  //.   - `$.Function ([a, b, a])` represents the `(a, b) -> a` type.
```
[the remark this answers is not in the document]
davidchambers: Yes. The behaviour of `$.Function` is unchanged. I believe sanctuary-def should be capable of expressing types of which we disapprove. ;)

index.js
davidchambers: Great point! I'll address this.

davidchambers: :zap:

index.js
davidchambers: This function references `$1`.

index.js
davidchambers: This function references `$1` and `$2`.

index.js
davidchambers: :zap:

test/index.js
davidchambers: I can imagine reformatting the source code too, but I'd prefer to do so in a separate pull request to avoid obfuscating the significant code changes.

test/index.js
davidchambers: I wanted to show that we're no longer limited to nine parameters. :joy:

index.js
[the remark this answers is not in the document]
davidchambers: I see a `:thumbsdown:` from Gabe. I certainly don't love the proposal. Let's vote!

  - :laughing: `$.env.concat([List ($.Unknown)])` ✘
  - :tada: `$.env.concat ([List ($.Unknown)])` ✔︎
  - :heart: `$.env .concat ([List ($.Unknown)])` ✔︎

The ✔︎/✘ indicates whether we'll be able to continue to use [`func-call-spacing`][1].


[1]: https://eslint.org/docs/rules/func-call-spacing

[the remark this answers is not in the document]
gabejohnson: I like the idea of the formatting as a signal that curried functions are being used.
```js
[1, 2, 3].concat([4, 5, 6]); // Nothing special going on here
S.concat ([1, 2, 3]) ([4, 5, 6]); // FP in the house!
```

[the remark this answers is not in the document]
davidchambers: I can see the logic behind that, Gabe. What about unary functions, though? `Math.abs(x)` or `Math.abs (x)`? `S.trim(s)` or `S.trim (s)`?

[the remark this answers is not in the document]
Avaq: What about `require('express')(b, c)`? What about `fs.reduce(f, x)(y)`? I think trying to categorise function calls based on underlying structure will raise more questions than it settles. I vote for the consistency of *always* putting a space between the function or method identifier:

* `require ('express') (a, b)`
* `fs.reduce (f, x) (y)`
* `$.env.concat ([List ($.Unknown)])`

index.js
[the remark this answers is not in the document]
Avaq: In Haskell `()` is a nullary tuple. I like to think that the only difference between Haskell and JavaScript, is that JavaScript forces you to pattern-match your tuples when passing them around, and you can only instantiate them at call time.

```js
//    f :: Tuple a b -> c -> ...
//      -- alernatively written as
//    f :: (a, b) -> c -> ...
const f = (a, b) => c => ...
```

When viewing it like this, Unit has no addressable value level member (only at call time).

index.js
Avaq: Oh, I missed that.

index.js
Avaq: Oh, I missed that.

index.js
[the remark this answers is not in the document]
davidchambers: `f(x)(y).z` is problematic. In `f (x) (y).z` the `.z` property access appears to be bound to `(y)` rather than to `f (x) (y)`. There are two reasonable solutions, each of which violates one of our current rules:

  - `f (x) (y) .z` (violates [`no-whitespace-before-property`][1])
  - `(f (x) (y)).z` (violates [`no-extra-parens`][2])

I prefer the latter as it is consistent with `foo.bar.baz` (I'm not ready to switch to `foo .bar .baz`).


[1]: https://eslint.org/docs/rules/no-whitespace-before-property
[2]: https://eslint.org/docs/rules/no-extra-parens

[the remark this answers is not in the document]
Avaq: So with this new approach, all preceding occurrences of `Z.env .concat` can be fixed, right?

index.js
```diff
@@ -2429,147 +2371,185 @@
...
-                                           index,
-                                           [],
-                                           [value]);
-            typeVarMap = assertRight(result).typeVarMap;
-          }
-          values[index] = value;
-        } else {
-          indexes.push(index);
+      return function(x) {
+        var args = slice.call(arguments);
```
Avaq: I think we might be able to avoid calling `slice.call` here. It would increase performance.

index.js
```diff
@@ -2429,147 +2371,185 @@
...
+            Right(typeVarMap) :
+          // else
+            Left(function() {
+              return invalidValue(env, typeInfo, index, propPath, x);
+            })
+        );
       }
+
+      var isThunk = expType.types.$1.type.type === NO_ARGUMENTS;
+      var numArgsExpected = isThunk ? 0 : expType.keys.length - 1;
```
Avaq: Note to self: I want to think about this approach.

index.js
```diff
@@ -2429,147 +2371,185 @@
...
-          tReprs.push(showType(typeInfo.types[idx]));
-        } else {
-          while (placeholders > 0) {
-            vReprs.push('__');
-            placeholders -= 1;
-          }
-          vReprs.push(Z.toString(_values[idx]));
+    var wrapped = typeInfo.types[0].type === NO_ARGUMENTS ?
+      function() {
+        var args = slice.call(arguments);
```
Avaq: Same here. We can move the `slice.call` part to the unhappy path.

index.js
davidchambers: Great point!

:zap:

index.js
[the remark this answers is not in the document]
davidchambers: :zap:

index.js
Avaq: My thoughts: https://github.com/sanctuary-js/sanctuary-def/issues/180

davidchambers:
I've rebased this pull request now that #184 has been merged. Please review __scripts/lint__ and __.eslint-es6.js__.

PR #182, opened 2018-02-13
types: rename Pair type constructor and define Array0 and Array1

davidchambers:
Commit message:

> Pair is a misleading name because a value of type `Pair a b` should support operations such as `map :: (b -> c) -> Pair a b -> Pair a c` and `bimap :: (a -> c) -> (b -> d) -> Pair a b -> Pair c d`.

The Sanctuary ecosystem will soon have a *real* Pair type. Let's free the name for use by that type. :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -473,6 +473,34 @@
   //. Constructor for homogeneous Array types.
   var Array_ = UnaryTypeWithUrl('Array', typeEq('Array'), id);
 
+  //# Array0 :: Type
+  //.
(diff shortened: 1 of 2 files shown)


Avaq:
Are we not going to use *Tuple* as a name?

davidchambers:
> Are we not going to use *Tuple* as a name?

What do you have in mind, exactly?

davidchambers:
The more I think about this change the more I like it. :)

PR #189, opened 2018-04-04
define Type#fantasy-land/equals

davidchambers:
I'd like to add a doctest for [`S.env`][1] such as this:

```javascript
//. > S.env
//. [ $.AnyFunction,
//. . $.Arguments,
//. . ...
//. . $.ValidDate,
//. . $.ValidNumber ]
```

In order to do so, two things must happen:

  - [`$.Type`][2] must be added to Sanctuary's default environment; and
  - `Type` values must satisfy the Setoid constraint imposed by [`S.equals`][3].

[1]: https://sanctuary.js.org/#env
[2]: https://github.com/sanctuary-js/sanctuary-def#Type
[3]: https://sanctuary.js.org/#equals

### DIFF
--- a/index.js
+++ b/index.js
@@ -361,6 +361,19 @@
 
   _Type['@@type'] = 'sanctuary-def/Type';
 
+  //  Type#fantasy-land/equals :: Type ~> Type -> Boolean
+  _Type.prototype['fantasy-land/equals'] = function(other) {
+    return (
+      Z.equals(this.type, other.type) &&
+      Z.equals(this.name, other.name) &&
+      Z.equals(this.url, other.url) &&
+      Z.equals(this.keys, other.keys) &&
+      this.keys.every(function(k) {
+        return Z.equals(this.types[k].type, other.types[k].type);
+      }, this)
+    );
+  };
+
   _Type.prototype.validate = function(x) {
     if (!this._test(x)) return Left({value: x, propPath: []});
     for (var idx = 0; idx < this.keys.length; idx += 1) {
--- a/test/index.js
+++ b/test/index.js
@@ -3242,3 +3242,26 @@ See https://github.com/sanctuary-js/sanctuary-def/tree/v${version}#Boolean for i
   });
 
 });
+
+suite ('interoperability', () => {
+
+  test ('Z.equals can operate on ‘Type’ values', () => {
+    eq (Z.equals ($.Number, $.Number)) (true);
+    eq (Z.equals ($.Number, $.String)) (false);
+    eq (Z.equals ($.Array ($.Number), $.Array ($.Number))) (true);
+    eq (Z.equals ($.Array ($.Number), $.Array ($.String))) (false);
+    eq (Z.equals ($.RecordType ({x: $.Number}), $.RecordType ({x: $.Number}))) (true);
+    eq (Z.equals ($.RecordType ({x: $.Number}), $.RecordType ({y: $.Number}))) (false);
+    eq (Z.equals ($.RecordType ({x: $.Number}), $.RecordType ({x: $.String}))) (false);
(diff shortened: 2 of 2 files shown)


test/index.js
```diff
@@ -3244,3 +3244,20 @@ See https://github.com/sanctuary-js/sanctuary-def/tree/v${version}#Boolean for i
...
+    eq (Z.equals ($.Number, $.Number)) (true);
+    eq (Z.equals ($.Number, $.String)) (false);
+    eq (Z.equals ($.Array ($.Number), $.Array ($.Number))) (true);
+    eq (Z.equals ($.Array ($.Number), $.Array ($.String))) (false);
+    eq (Z.equals ($.RecordType ({x: $.Number}), $.RecordType ({x: $.Number}))) (true);
+    eq (Z.equals ($.RecordType ({x: $.Number}), $.RecordType ({y: $.Number}))) (false);
+    eq (Z.equals ($.RecordType ({x: $.Number}), $.RecordType ({x: $.String}))) (false);
+    eq (Z.equals ($.NullaryType ('X') ('') (x => true), $.NullaryType ('X') ('') (x => true))) (true);
+    eq (Z.equals ($.NullaryType ('X') ('') (x => true), $.NullaryType ('Y') ('') (x => true))) (false);
+    eq (Z.equals ($.NullaryType ('X') ('') (x => true), $.NullaryType ('X') ('') (x => false))) (true);
```
Avaq: Is this something we want?

index.js
```diff
@@ -361,6 +361,14 @@
 
   _Type['@@type'] = 'sanctuary-def/Type';
 
+  //  Type#fantasy-land/equals :: Type ~> Type -> Boolean
+  _Type.prototype['fantasy-land/equals'] = function(other) {
+    return Z.equals(this.type, other.type) &&
+           Z.equals(this.name, other.name) &&
+           Z.equals(this.url, other.url) &&
+           Z.equals(Z.toString(this), Z.toString(other));
```
Avaq: Do I understand correctly that we rely on the behaviour of `toString` for asserting equality of nested types?

test/index.js
davidchambers: It's documenting the fact that we don't attempt to determine function equivalence. Relying on function identity doesn't feel good, and shouldn't be necessary given that the `url` field alone should discriminate between identically named type constructors.

index.js
davidchambers: It's a simple approach but I now see that it ignores the `url` fields of nested types, leading to avoidable collisions between identically named types. I'll rework this.

Avaq: That. :)

index.js
```diff
@@ -363,10 +363,15 @@
...
-           Z.equals(this.name, other.name) &&
-           Z.equals(this.url, other.url) &&
-           Z.equals(Z.toString(this), Z.toString(other));
+    return (
+      Z.equals(this.type, other.type) &&
+      Z.equals(this.name, other.name) &&
+      Z.equals(this.url, other.url) &&
+      Z.equals(this.keys, other.keys) &&
+      this.keys.every(function(k) {
+        return Z.equals(this.types[k].type, other.types[k].type);
```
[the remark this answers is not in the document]
Avaq: I was worried that `this.types[n]` might lead back to a type we're currently processing in the stack, leading to infinite recursion. Based on your example I now see that this can't happen. The `this.types` refers only to the *type* parameters and **not** to the *constructor* parameters.

test/index.js
ivenmarquardt: I just checked my notifications and became aware of this conversation. I recently added a `Ref` type to my lib, which mainly implements `Setoid` to allow reference equality checks. Doing this in a type-safe and explicit manner is sometimes useful, e.g. for change detection of persistent data structures. Just thought it might interest you.

PR #192, opened 2018-04-13
underline function argument return types precisely

### LINKED ISSUE #181: imprecise underlining upon function argument returning invalid value
If the return value of a function *argument* is incorrect, the entire return type is underlined even if the outermost type is correct. For example:

```javascript
S.unfoldr (c => c <= 'z' ? S.Just (c) : S.Nothing) ('a');
// ! TypeError: Invalid value
//
//   unfoldr :: (b -> Maybe (Pair a b)) -> b -> Array a
//                    ^^^^^^^^^^^^^^^^
//                           1
//
//   1)  Just("a") :: Maybe String
//
//   The value at position 1 is not a member of ‘Maybe (Pair a b)’.
//
//   See https://github.com/sanctuary-js/sanctuary/tree/v0.14.1#MaybeType for information about the sanctuary/Maybe type.
```

Only `Pair a b` should be underlined in this case.

davidchambers:
Closes #181

The updated code is simpler in addition to producing better error messages. :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -273,13 +273,6 @@
   //  or :: (Array a, Array a) -> Array a
   function or(xs, ys) { return isEmpty(xs) ? ys : xs; }
 
-  //  singleton :: (String, a) -> StrMap a
-  function singleton(k, v) {
-    var result = {};
-    result[k] = v;
-    return result;
-  }
-
   //  strRepeat :: (String, Integer) -> String
   function strRepeat(s, times) {
     return Array(times + 1).join(s);
@@ -1096,7 +1089,7 @@
           return invalidValue(env,
                               typeInfo,
                               index,
-                              result.value.propPath,
+                              Z.concat(propPath, result.value.propPath),
                               result.value.value);
         });
       }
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1147,49 +1140,27 @@
...
-                                   r.typeVarMap,
-                                   expType.types[k].type,
-                                   index,
-                                   Z.concat(propPath, [k]),
-                                   innerValues);
-                    },
-                    Z.reduce(function(e, x) {
-                      var t = expType.types[k].type;
-                      return Z.chain(function(r) {
-                        return test(env)(t)(x) ? Right(r) : Left(function() {
```
davidchambers: This was the problem. `test(env)(t)(x)` expands to `satisfactoryTypes(env, {name: 'name', constraints: {}, types: [t]}, {}, t, 0, [], [x]).isRight`, which throws away useful information in the failure case. This forced us to infer lost context below (see the uses of `typeVarConstraintViolation` and `invalidValue`).

Issue #212, opened 2018-07-01
Smaller stack traces


danielo515:
Hello,

I know there is already an issue about cleaner stack traces. This is somewhat related, but focused on the length, and not specifically on the number of stacks or where they come from.

Here is a small example of one type error I got today:

```
$ node test                                                                                                                            ‹ruby-2.3.0›
/Users/danielo/GIT/danielo-packages/packages/fp-min/node_modules/sanctuary-def/index.js:2513
          throw invalidArgumentsCount (typeInfo, index, 1, args);
          ^

TypeError: ‘sanfu/promise/pipeAcc’ applied to the wrong number of arguments

sanfu/promise/pipeAcc :: Array Function -> Function
                         ^^^^^^^^^^^^^^
                               2

Expected one argument but received three arguments:

  - function(x) {
        var args = slice.call (arguments);
        if (args.length !== 1) {
          throw invalidArgumentsCount (typeInfo, index, 1, args);
        }
        var typeVarMap = (assertRight (
          satisfactoryTypes (env,
                             typeInfo,
                             _typeVarMap,
                             typeInfo.types[index],
                             index,
                             [],
                             args)
        )).typeVarMap;

        var values = Z.concat (_values, args);
        if (index + 1 === n) {
          var value = values.reduce (function(f, x, idx) {
            return f (wrapFunctionCond (typeVarMap, idx, x));
          }, impl);
          typeVarMap = (assertRight (
            satisfactoryTypes (env,
                               typeInfo,
                               typeVarMap,
                               typeInfo.types[n],
                               n,
                               [],
                               [value])
          )).typeVarMap;
          return wrapFunctionCond (typeVarMap, n, value);
        } else {
          return wrapNext (typeVarMap, values, index + 1);
        }
      }
  - ([, b]) => b * 2
  - ([, , c]) => c - 1

    at invalidArgumentsCount (/Users/danielo/GIT/danielo-packages/packages/fp-min/node_modules/sanctuary-def/index.js:2069:12)
    at /Users/danielo/GIT/danielo-packages/packages/fp-min/node_modules/sanctuary-def/index.js:2513:17
    at Object.<anonymous> (/Users/danielo/GIT/danielo-packages/packages/fp-min/test/index.js:4:11)
    at Module._compile (internal/modules/cjs/loader.js:702:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:713:10)
[...]

Issue #64, opened 2016-05-06
Cleaner stack traces


jwoudenberg:
I was wondering what people think of using a lib like [stack-utils](https://github.com/tapjs/stack-utils) to clean up the stack trace to ensure the first line of the stack corresponds to the function call that causes the TypeError.

davidchambers:
Could you provide an example of some code containing a type error, the current stack trace, and the stack trace we'd see instead were we to use stack-utils?

jwoudenberg:
Sure! Take the following code with improperly uses Sanctuary's maybe:

```js
const { Just } = require('sanctuary')
Just(4).chain(x => x)
```

It results in the following error:

```
/home/jasper/tmp/experiments/sanctuary/node_modules/sanctuary-def/index.js:1268
              throw invalidValue(name,
              ^

TypeError: Invalid value

Maybe#chain :: Maybe a -> Function -> Maybe b
                                      ^^^^^^^
                                         1

1)  4 :: Number, FiniteNumber, NonZeroFiniteNumber, Integer, ValidNumber

The value at position 1 is not a member of ‘Maybe b’.

    at invalidValue (/home/jasper/tmp/experiments/sanctuary/node_modules/sanctuary-def/index.js:988:12)
    at /home/jasper/tmp/experiments/sanctuary/node_modules/sanctuary-def/index.js:1268:21
    at /home/jasper/tmp/experiments/sanctuary/node_modules/sanctuary-def/index.js:571:20
    at apply (/home/jasper/tmp/experiments/sanctuary/node_modules/ramda/dist/ramda.js:1184:19)
    at Object.f2 [as apply] (/home/jasper/tmp/experiments/sanctuary/node_modules/ramda/dist/ramda.js:473:22)
    at Maybe.<anonymous> (/home/jasper/tmp/experiments/sanctuary/node_modules/sanctuary/index.js:353:38)
    at Maybe.<anonymous> (/home/jasper/tmp/experiments/sanctuary/node_modules/sanctuary-def/index.js:1264:34)
    at Maybe.chain (/home/jasper/tmp/experiments/sanctuary/node_modules/sanctuary-def/index.js:567:20)
    at Object.<anonymous> (/home/jasper/tmp/experiments/sanctuary/index.js:2:9)
    at Module._compile (module.js:541:32)
```

[...]

PR #213, opened 2018-07-01
WIP: Exclude internal stack frames from littering error messages

Avaq:
Opening the PR because I'm not sure when I'll get back to it, so it can serve as an example.

These changes show an approach to tackling issues like #64 and (maybe) #212 using `Error.captureStacktrace` from V8 to trim the internal stack frames from the stack trace present on instances of Error.

### DIFF
--- a/index.js
+++ b/index.js
@@ -239,6 +239,9 @@
   //  always2 :: a -> (b, c) -> a
   function always2(x) { return function(y, z) { return x; }; }
 
+  //  captureStackTrace :: Assignable a => (a, Function) -> a
+  var captureStackTrace = Error.captureStackTrace || noop;
+
   //  compose :: (b -> c, a -> b) -> (a -> c)
   function compose(f, g) {
     return function(x) {
@@ -280,6 +283,9 @@
       return xs.some (function(x) { return Z.equals (x, y); });
     };
   }
+
+  //  noop :: () -> Undefined
+  function noop() {}
 
   //  or :: (Array a, Array a) -> Array a
   function or(xs, ys) { return isEmpty (xs) ? ys : xs; }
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -2061,12 +2067,12 @@
   //. Multiple constraints may be placed on a type variable by including
   //. multiple `TypeClass` values in the array (e.g. `{a: [Foo, Bar, Baz]}`).
 
-  //  invalidArgumentsCount :: (TypeInfo, Integer, Integer, Array Any) -> Error
+  //  invalidArgumentsCount :: (TypeInfo, Integer, Integer, Array Any, Function) -> Error
```
Avaq: Note: This line is too long now. Argument types should be moved into the argument list as is done with other functions.

index.js
```diff
@@ -239,6 +239,9 @@
   //  always2 :: a -> (b, c) -> a
   function always2(x) { return function(y, z) { return x; }; }
 
+  //  captureStackTrace :: Assignable a => (a, Function) -> a
+  var captureStackTrace = Error.captureStackTrace || noop;
```
davidchambers: Should the return type be `Undefined` rather than `a`? As written `noop` should not be used here because it does not return its first argument.

Avaq: Oh, you're right. captureStackTrace does actually return `undefined`.

davidchambers:
I would love to see you revive this pull request, Aldwin. :)

Avaq:
I probably will. I am aware if its existence, and it's one of many things on my ever-evolving to-do list. :P
I'm trying out a thing in my latest rework of Fluture, which, if successful, might make its way into a small library that we can base this PR on.

Issue #138, opened 2017-04-28
Named Record Types


JAForbes:
I work with a lot of objects, it'd be helpful to tag them with a name.  It'd make for more terse signatures.

I still think we should show the complete interface if a type error occurs though.  But when rendering the signature, it'd be preferable to just use the name, I'm finding type errors nearly always wrap the terminal/repl because I'm working with large records.

Then the `^^^^` don't line up with the particular argument, so its not immediately obvious what is causing it to fail.

davidchambers:
Good idea.

An idea that just occurred to me is to have the representation span multiply lines:

```
{
  bar :: Array String
  baz :: {
    quux :: Integer
            ^^^^^^^
               1
  }
  foo :: Boolean
}
```

As opposed to:

```
{ bar :: Array String, baz :: { quux :: Integer }, foo :: Boolean }
                                        ^^^^^^^
                                           1
```

We could use the name of the record type in the signature, then expand the type in the description below and underline the invalid portion.

JAForbes:
That'd be fantastic!

kurtmilam:
It'd be great to have the representation span multiple lines. I have some records that wrap even on a 4K monitor :D

JAForbes:
Some new thoughts: 

I'm using this to validate API responses and there are a lot of fields, sometimes its hard to see what is wrong with a value when there are 50 keys with various levels of nesting.

It's not that useful to render the complete type for large objects (say keys > 5, or levels of nesting greater than 2).  It'd be a lot more useful to me to list the paths and types that were invalid, with the expected type and the actual value.  Particularly when there are multiple validation issues.

```
TypeError ... yada

1. Value -123 at path: a.b.c.d is not a member of PositiveNumber
2. Value [1,2,3] at path: a.b.c.e is not a member of List String
```

It could also be handy to show the values in context, with the valid values filtered out.

```
TypeError Provide value did not satisfy custom RecordType API Response

The provided value did not satisfy the following Type constraints:

Preview:

  { a: 
    { b: 
      { c: 
         { d: -123, e: [1,2,3] }, ... 
       }, ... 
     }, ... 
  } 

1. Value -123 at path: a.b.c.d is not a member of PositiveNumber
2. Value [1,2,3] at path: a.b.c.e is not a member of List String
```

It'd be great too if validate, or something similar could provide this metadata as an object of the form: `{ preview :: String, failures:: List { value:: Any, path: List String, type: Type } }`

[...]

PR #236, opened 2019-03-26
support explicit subtyping

davidchambers:
Before:

```javascript
var FiniteNumber = $.NullaryType
  ('FiniteNumber')
  ('https://github.com/sanctuary-js/sanctuary-def/tree/v0.19.0#FiniteNumber')
  (function(x) { return ValidNumber._test (x) && isFinite (x); });
```

After:

```javascript
var FiniteNumber = $.NullaryType
  ('FiniteNumber')
  ('https://github.com/sanctuary-js/sanctuary-def/tree/v0.19.0#FiniteNumber')
  ([ValidNumber])
  (isFinite);
```

The primary reason for this change is to make defining subtypes more convenient. In the future the relationships this change exposes could be used to make type checking more efficient. If a value is not a member of `$.Number`, say, we know it is not a member of any of the subtypes of `$.Number` either.

### DIFF
--- a/index.js
+++ b/index.js
@@ -251,6 +251,9 @@
   //  always2 :: a -> (b, c) -> a
   function always2(x) { return function(y, z) { return x; }; }
 
+  //  complement :: (a -> Boolean) -> a -> Boolean
+  function complement(pred) { return function(x) { return !(pred (x)); }; }
+
   //  init :: Array a -> Array a
   function init(xs) { return xs.slice (0, -1); }
 
@@ -285,6 +288,12 @@
 
   //  or :: (Array a, Array a) -> Array a
   function or(xs, ys) { return isEmpty (xs) ? ys : xs; }
+
+  //  prop :: String -> {} -> a
+  function prop(field) { return function(record) { return record[field]; }; }
+
+  //  sizeEq :: Foldable f => Integer -> f a -> Boolean
+  function sizeEq(n) { return function(xs) { return Z.size (xs) === n; }; }
 
   //  strRepeat :: (String, Integer) -> String
   function strRepeat(s, times) {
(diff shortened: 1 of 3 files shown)


davidchambers:
:warning: Before this pull request is merged, we must decide what to do with record types.

Currently, record types are shown in `{ name :: type, name :: type }` form in type signatures and error messages. In some cases this makes it possible to indicate which field is incorrectly typed.

Record types can now subtype other record types. We could make code changes to ensure that `$.RecordType ([$.RecordType ([]) ({x: X})]) ({y: Y})` behaves identically to `$.RecordType ([]) ({x: X, y: Y})`. This special case would not apply to record types defined as subtypes of non-record types. This is an edge case, admittedly, but imagine reading an error message stating that `{y: 0}` is not a member of `{ y :: Number }` (due to an unacknowledged dependency). Confusing.

One option is to leave `$.RecordType` as it exists on `master`, only capable of defining anonymous record types without supertypes. We would add a second type constructor, `$.NamedRecordType`, that would also take `name`, `url`, and `supertypes`. Types defined via this constructor would be referred to by name in type signatures and error messages, avoiding the problem @JAForbes noted in #138.

Avaq:
This seems like the first in a series of steps towards the overly ambitious goals I set for myself with #162. Maybe it's time to close that PR, seeing as the branch has fallen quite far behind, and incremental steps probably will get us there faster.

davidchambers:
Aha! I thought we had discussed this. I searched issues but not pull requests, so failed to find #162.

davidchambers:
Let's merge #238 before taking further action here.

davidchambers:
This pull request is now ready to be reviewed. :)

index.js
```diff
@@ -1677,6 +1761,8 @@
   //.
   //. To define an anonymous record type one must provide:
   //.
+  //.   - an array of supertypes (exposed as `t.supertypes`); and
+  //.
```
davidchambers: This was an erroneous inclusion. I will remove it.

davidchambers: #254

PR #240, opened 2019-03-28
types: require records to be object-like

davidchambers:
This pull request corrects the surprising semantics documented in #239.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -1007,11 +1007,11 @@ Since there is no type of which all the above values are members, the type-varia
     //    Empty :: Type
     const Empty = $.RecordType ({});
 
-    eq ($.test ([]) (Empty) (null)) (true);
-    eq ($.test ([]) (Empty) (undefined)) (true);
-    eq ($.test ([]) (Empty) (false)) (true);
-    eq ($.test ([]) (Empty) (12.34)) (true);
-    eq ($.test ([]) (Empty) ('xyz')) (true);
+    eq ($.test ([]) (Empty) (null)) (false);
+    eq ($.test ([]) (Empty) (undefined)) (false);
+    eq ($.test ([]) (Empty) (false)) (false);
+    eq ($.test ([]) (Empty) (12.34)) (false);
+    eq ($.test ([]) (Empty) ('xyz')) (false);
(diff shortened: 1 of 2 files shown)


test/index.js
```diff
@@ -1007,11 +1007,11 @@ Since there is no type of which all the above values are members, the type-varia
...
-    eq ($.test ([]) (Empty) (null)) (true);
-    eq ($.test ([]) (Empty) (undefined)) (true);
-    eq ($.test ([]) (Empty) (false)) (true);
-    eq ($.test ([]) (Empty) (12.34)) (true);
-    eq ($.test ([]) (Empty) ('xyz')) (true);
+    eq ($.test ([]) (Empty) (null)) (false);
+    eq ($.test ([]) (Empty) (undefined)) (false);
+    eq ($.test ([]) (Empty) (false)) (false);
+    eq ($.test ([]) (Empty) (12.34)) (false);
+    eq ($.test ([]) (Empty) ('xyz')) (false);
```
davidchambers: I'm convinced that `null` and `undefined` should not be considered members of `{}`, the empty record type, but I'm unsure about these other changes. If `'xyz'` is not a member of `{}`, it must not be a member of `{ length :: Integer }` either, which seems unfortunate.

Perhaps the guard should be `x != null` rather than `Object (x) === x`?

Avaq: Besides feeling a little bit of unease thinking about using RecordType to define the "virtual shape" of scalar values, I don't see it causing *actual* problems. A *hypothetical* problem would be that some unexpected scalar value coincidentally matches some model within my domain, but I can't think of a concrete example.

The benefit is being able to define commonalities between the types of scalar values by using RecordType. That seems neat, but I'm also not sure how beneficial that is given that we can do it with a simple NullaryType.

All in all the pros and cons are both weak, and weight very lightly to create little significant disbalance. In summary: I don't have a particularly strong opinion about it.

davidchambers: I agree that the pros and cons are both weak. This being the case, I suggest that we take the smaller of the two possible steps. We are free to take the second step in the future if we decide that `'xyz'` should not be a member of `{ length :: Integer }`.

I will close this pull request and open a smaller one in its place.

PR #255, opened 2019-04-03
provide keys, extractors, and types to ‘_Type’ as one data structure

davidchambers:
Commit message:

> This commit changes where extractor functions and inner types are stored on Type objects. This is technically a breaking change, because prior to this commit the documentation divulged the locations of inner types.

### DIFF
--- a/index.js
+++ b/index.js
@@ -254,6 +254,13 @@
   //  complement :: (a -> Boolean) -> a -> Boolean
   function complement(pred) { return function(x) { return !(pred (x)); }; }
 
+  //  filter :: Filterable f => (a -> Boolean) -> f a -> f a
+  function filter(pred) {
+    return function(xs) {
+      return Z.filter (pred, xs);
+    };
+  }
+
   //  init :: Array a -> Array a
   function init(xs) { return xs.slice (0, -1); }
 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -417,17 +422,23 @@
     format,     // :: (String -> String, String -> String -> String) -> String
     supertypes, // :: Array Type
     test,       // :: Any -> Boolean
-    keys,       // :: Array String
-    types       // :: StrMap { extractor :: a -> Array b, type :: Type }
+    tuples      // :: Array (Array3 String (a -> Array b) Type)
```
davidchambers: Currently we take `keys` and `types`, and assume that the elements of `keys` correspond to the keys of `types`. With this change it's not possible to have incoherent data. :)

index.js
```diff
@@ -2643,7 +2636,7 @@
         );
       }
 
-      var isThunk = expType.types.$1.type.type === NO_ARGUMENTS;
+      var isThunk = expType.types.$1.type === NO_ARGUMENTS;
```
davidchambers: Removing occurrences of `.type.type` can only be a good thing. :rofl:

index.js
```diff
@@ -1070,17 +1086,18 @@
       } else {
         seen$ = seen;
       }
+      var expandUnknown2 = expandUnknown4 (seen$) (value);
```
davidchambers: When partially applying curried functions for the sake of efficiency, I've been finding it helpful to indicate the number of arguments remaining. It's difficult to come up with unique names without such a system. :)

PR #256, opened 2019-04-03
allow extractors to return any foldable structure

davidchambers:
I realized while working on [sanctuary-list][1] that we need not require an *array* of values specifically. Many algebraic data types are foldable, so in some cases the identity function should be able to serve as an extractor function.

I tried generalizing the code to avoid using `toArray` anywhere, but this proved impractical due to uses of `Z.chain` with functions that always return arrays. The solution is to compose each of the extractors with `toArray`, necessitating very few internal changes.

[1]: https://github.com/sanctuary-js/sanctuary-list

### DIFF
--- a/index.js
+++ b/index.js
@@ -264,8 +264,8 @@
   //  init :: Array a -> Array a
   function init(xs) { return xs.slice (0, -1); }
 
-  //  isEmpty :: Array a -> Boolean
-  function isEmpty(xs) { return xs.length === 0; }
+  //  isEmpty :: Foldable f => f a -> Boolean
+  function isEmpty(xs) { return Z.size (xs) === 0; }
 
   //  isPrefix :: Array a -> Array a -> Boolean
   function isPrefix(candidate) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -685,7 +693,7 @@
     ('Maybe')
     ([])
     (typeEq ('sanctuary-maybe/Maybe@1'))
-    (function(maybe) { return maybe.isJust ? [maybe.value] : []; });
+    (I);
```
davidchambers: :tada:

index.js
```diff
@@ -909,11 +917,7 @@
     ('StrMap')
     ([Object_])
     (K (true))
-    (function(strMap) {
-       return Z.reduce (function(xs, x) { xs.push (x); return xs; },
-                        [],
-                        strMap);
-     });
+    (I);
```
davidchambers: :tada:

index.js
```diff
@@ -1605,10 +1609,10 @@
                      (t.url)
                      (t.supertypes)
                      (t._test ([]))
-                     (t.extractors.$1);
+                     (t._extractors.$1);
```
davidchambers: We provide the user-provided extractor function here to avoid `B (toArray) (B (toArray) (_1))`.

PR #264, opened 2019-04-06
simplify code for formatting type signatures

davidchambers:
Prior to this change, parameterized types would be formatted with enclosing parens. `$.Array (a)`, for example, would be formatted as `'(Array a)'`. In many contexts these parens could safely be stripped, but doing so required convoluted conditional logic:

```javascript
return unless (
  t.type === FUNCTION || t.type === RECORD || isEmpty (t.keys),
  stripOutermostParens,
  ...
);
```

```javascript
return unless (t.type === RECORD ||
                 isEmpty (t.keys) ||
                 t.type === FUNCTION && isEmpty (propPath) ||
                 !(isEmpty (propPath)),
               stripOutermostParens,
               ...);
```

```javascript
var repr = show (type);
var parenthesized = repr.slice (0, 1) + repr.slice (-1) === '()';
return formatType4 (function(s) {
  return parenthesized && repr !== '()' && s.length === repr.length ?
    _ ('(') + r ('^') (s.slice (1, -1)) + _ (')') :
    r ('^') (s);
});
```

This pull request redefines a responsibility of parameterized types in a small but impactful way: rather than *stripping* parens *if possible*, parameterized types now *wrap* inner types in parens *if necessary*.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -1456,7 +1456,7 @@ Since there is no type of which all the above values are members, the type-varia
     eq (typeof $.Array1) ('function');
     eq ($.Array1.length) (1);
     eq (show ($.Array1)) ('Array1 :: Type -> Type');
-    eq (show ($.Array1 (a))) ('(Array1 a)');
+    eq (show ($.Array1 (a))) ('Array1 a');
     eq (($.Array1 (a)).name) ('Array1');
     eq (($.Array1 (a)).url) (`https://github.com/sanctuary-js/sanctuary-def/tree/v${version}#Array1`);
     eq (($.Array1 (a)).supertypes) ([$.Array ($.Unknown)]);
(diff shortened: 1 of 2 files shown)


test/index.js
```diff
@@ -1556,10 +1556,36 @@ See https://github.com/sanctuary-js/sanctuary-def/tree/v${version}#Array2 for in
...
-    eq (show ($.Function ([a, b]))) ('(a -> b)');
+    eq (show ($.Function ([a, b]))) ('a -> b');
     eq (($.Function ([a, b])).name) ('');
     eq (($.Function ([a, b])).url) ('');
     eq (($.Function ([a, b])).supertypes) ([$.AnyFunction]);
+
+    //  Fn :: Type -> Type -> Type
+    function Fn($1) { return function($2) { return $.Function ([$1, $2]); }; }
+
+    //  https://gist.github.com/Avaq/1f0636ec5c8d6aed2e45
```
Avaq: :D

PR #265, opened 2019-04-07
track type arity explicitly rather than inferring it from ‘keys’

davidchambers:


index.js
```diff
@@ -434,12 +435,13 @@
...
       return _extractors;
     }, {});
+    t.arity = arity;  // number of type parameters
     t.extractors = Z.map (B (toArray), t._extractors);
     t.format = format || function(outer, inner) {
       return Z.reduce (function(s, tuple) {
         return s +
                outer (' ') +
-               when (tuple[2].keys.length > 0 && tuple[2].type !== RECORD)
+               when (tuple[2].arity > 0)
```
davidchambers: The problem with using `type.keys.length` is that record types can have many keys but should *not* be considered parameterized types. By using `type.arity` we need not remember to exclude record types.

index.js
```diff
@@ -1211,12 +1206,12 @@
       $typeVarMap[typeVar.name].valuesByPath[key].push (value);
       $typeVarMap[typeVar.name].types = Z.chain (function(t) {
         return (
-          t.keys.length < typeVar.keys.length || !(isValid (t) (value)) ?
+          !(isValid (t) (value)) ?
```
davidchambers: This deletion could have been part of #250.

PR #280, opened 2020-01-08
types: define Void

davidchambers:
I struggled with the description, and I'm not completely satisfied with the first sentence of the second paragraph. Suggestions are welcome. :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -531,6 +531,18 @@
   var Unknown =
   _Type (UNKNOWN, '', '', 0, always2 ('Unknown'), [], K (K (true)), []);
 
+  //# Void :: Type
+  //.
+  //. Uninhabited type.
+  //.
+  //. May be used to convey that a type parameter of an algebraic data type
+  //. will not be used. For example, a future of type `Future Void String`
+  //. will never be rejected.
+  var Void = NullaryTypeWithUrl
+    ('Void')
+    ([])
+    (K (false));
+
   //# Any :: Type
   //.
   //. Type comprising every JavaScript value.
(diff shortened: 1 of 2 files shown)


Avaq:
What's the motivation behind the addition of this type?

davidchambers:
> What's the motivation behind the addition of this type?

While @voxbono and I were working on a [functional server][1] we decided to use futures for HTTP responses. We then realized we had no use for the first type parameter: even if the database was inaccessible we could *resolve* with a `Response` value for a server error. I was about to suggest using `Future Null Response` but realized that `Future Void Response` would convey more information.


[1]: https://github.com/voxbono/functional-server

Avaq:
Hmm, usually `Future a Response` conveys just as much information - it is assumed that a completely unrestricted type variable has no concrete type.

The problem with `Future a Response` is that `reject (42)` is a valid member of it from the perspective of the type checker.

So you want a type that says "the type at this position must never become concrete". Maybe the description could be something like:

*Indicates that the type will never become concrete. For example, a future of type `Future Void String` will never be rejected.*

davidchambers:
> _Indicates that the type will never become concrete._

I'm not completely satisfied with this sentence either. I'm struggling to use a *type* to indicate that no *value* will ever be present at the position in question as I lack terminology for associating values with type parameters.

Avaq:
> the type at **this position** must never become concrete

> will ever be present at **the position** in question


I guess we both think of it as "position". Maybe we can use this word?

Avaq:
When talking in terms of "concreteness" we can refer to the type, however.

davidchambers:
I like “position”. The current description refers to position, albeit indirectly:

> May be used to convey that a type parameter of an algebraic data type will not be used.

This, for example, refers to the positional arguments `a` and `b` in `Future a b`.

I'll merge this pull request in its current form, but I'd like to improve the description in the future.

PR #281, opened 2020-01-08
types: define JsSet and JsMap type constructors

### LINKED ISSUE #183: Set, WeakSet, Map and WeakMap should be in the default env
I believe that these built-in collection types should be included in the default `env`

davidchambers:
Closes #183

I set out to define `$.JsWeakSet` and `$.JsWeakMap` as well, but I decided not to do so when I learnt how strange [`WeakSet`][1] and [`WeakMap`][2] are. Users can define (nullary) types for these if necessary.

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakSet
[2]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap

### DIFF
--- a/index.js
+++ b/index.js
@@ -710,6 +710,21 @@
     (typeEq ('sanctuary-identity/Identity@1'))
     (I);
 
+  //# JsMap :: Type -> Type -> Type
+  var JsMap = BinaryTypeWithUrl
+    ('JsMap')
+    ([])
+    (function(x) { return toString.call (x) === '[object Map]'; })
+    (function(jsMap) { return Array.from (jsMap.keys ()); })
+    (function(jsMap) { return Array.from (jsMap.values ()); });
+
+  //# JsSet :: Type -> Type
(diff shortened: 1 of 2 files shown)


Avaq:
I'm not sure why you would leave out the weak variants. Is it to discourage usage of them?

davidchambers:
Weak sets can only contain objects, so the type would be `JsWeakSet` rather than `JsWeakSet a`. Similarly, weak maps can only contain objects as keys, so the type would be `JsWeakMap a` rather than `JsWeakMap a b`. The values of a weak map are not enumerable, though, so the extractor would have to be `jsWeakMap => []`.

These quirks deterred me, but if you think we should define `$.JsWeakSet` and `$.JsWeakMap` I will do so in a separate pull request. :)

Avaq:
I see. Sounds good to me. Let's leave the weak variants to be defined by users. We can always make another PR as you said.

PR #282, opened 2020-01-09
types: document JsSet and JsMap type constructors

davidchambers:
I forgot to include descriptions in #281.


davidchambers:
When reviewing pull requests one can scrutinize additions; omissions are easily overlooked. :mag:

PR #292, opened 2020-04-30
Readme fix (index.js)

mreinigjr:
I made an attempt to better clarify the object accepted by the `create` method, the arguments accepted by the returned function (`def`) from the `create` method, added type signatures for both `create` and `def` and commented an example where `def` is used.  All changes are simple and my fp knowledge is limited, but growing :smile:. So please let me what is wrong or feel free to correct anything that is not documented correctly.  Lastly, I could not figure out how you auto generated the docs from the index.js file, so the end result may not be quite right, but all tests are passing.

### DIFF
--- a/index.js
+++ b/index.js
@@ -61,9 +61,11 @@
 //. const env = $.env.concat ([List ($.Unknown)]);
 //. ```
 //.
-//. The next step is to define a `def` function for the environment:
+//. The next step is to define a `def` function for the environment using
+//. `$.create`:
 //.
 //. ```javascript
+//. //    def :: String -> StrMap (Array TypeClass) -> Array Type -> Function -> Function
 //. const def = $.create ({checkTypes: true, env});
 //. ```
 //.
@@ -72,6 +74,7 @@
 //. during development. For example:
 //.
 //. ```javascript
+//. //    def :: String -> StrMap (Array TypeClass) -> Array Type -> Function -> Function
 //. const def = $.create ({
 //.   checkTypes: process.env.NODE_ENV === 'development',
 //.   env,
@@ -83,10 +86,10 @@
 //. ```javascript
 //. //    add :: Number -> Number -> Number
 //. const add =
-//. def ('add')
-//.     ({})
-//.     ([$.Number, $.Number, $.Number])
-//.     (x => y => x + y);
+//. def ('add')                           // name
+//.     ({})                              // type-class constraints
+//.     ([$.Number, $.Number, $.Number])  // input and output types
+//.     (x => y => x + y);                // implementation
 //. ```
 //.
 //. `[$.Number, $.Number, $.Number]` specifies that `add` takes two arguments


davidchambers:
Thanks for submitting this pull request, @mreinigjr! I will leave comments in situ.

index.js
```diff
@@ -63,8 +63,13 @@
...
 //. The next step is to define a `def` function for the environment:
 //.
+//# create :: { checkTypes :: Boolean, env :: Array Type } -> Module
+//.
 //. ```javascript
-//. const def = $.create ({checkTypes: true, env});
+//. const def = $.create ({
+//.   checkTypes: true,
+//.   env: env,
+//. });
```
davidchambers, line 72: `{checkTypes: true, env}` is equivalent to `{checkTypes: true, env: env}` in modern environments, so I don't think this change is necessary.

index.js
```diff
@@ -63,8 +63,13 @@
 //.
 //. The next step is to define a `def` function for the environment:
 //.
+//# create :: { checkTypes :: Boolean, env :: Array Type } -> Module
```
davidchambers, line 66:
```suggestion
//# create :: { checkTypes :: Boolean, env :: Array Type } -> String -> StrMap (Array TypeClass) -> Array Type -> Function -> Function
```

index.js
```diff
@@ -74,19 +79,24 @@
...
-//.   env,
+//.   env: env,
 //. });
 //. ```
 //.
-//. `def` is a function for defining functions. For example:
+//. `def` is a function for defining functions.
+//.
+//# def :: String -> StrMap Array TypeClass -> Array Type -> Function
+//. -> Function
```
davidchambers, line 89: I don't think we should use `//#` here, as `def` is not one of the module's exports. We could, though, include a type signature in the preceding code block:

```diff
 //. ```javascript
+//. //    def :: String -> StrMap (Array TypeClass) -> Array Type -> Function -> Function
 //. const def = $.create ({
 //.   checkTypes: process.env.NODE_ENV === 'development',
 //.   env,
 //. });
 //. ```
```

index.js
```diff
@@ -63,8 +63,13 @@
 //.
 //. The next step is to define a `def` function for the environment:
```
davidchambers, line 64: This line has been separated from the code block meant to follow it by the heading below. Let's think of a way to provide API documentation for `$.create` without interrupting the flow of this tutorial.

index.js
```diff
@@ -74,19 +79,24 @@
...
 //.
 //. ```javascript
 //. //    add :: Number -> Number -> Number
 //. const add =
-//. def ('add')
-//.     ({})
-//.     ([$.Number, $.Number, $.Number])
-//.     (x => y => x + y);
+//. def ('add') // name of function
+//.     ({}) // typeclass constraints
```
davidchambers, line 97:
```suggestion
//.     ({}) // type-class constraints
```

index.js
mreinigjr: sounds good.

index.js
mreinigjr: updated.

index.js
```diff
@@ -74,19 +79,24 @@
 //. ```javascript
 //. const def = $.create ({
 //.   checkTypes: process.env.NODE_ENV === 'development',
-//.   env,
+//.   env: env,
```
mreinigjr, line 82: I reverted this per your previous statement.

index.js
mreinigjr: updated.

index.js
mreinigjr: updated.

index.js
mreinigjr: After thinking this over some more, I don't really think documenting the type signature of the `create` method has nearly the same value as documenting the type signature of the `def` module.  I completely agree with your comment and I think we should just revert it back.  I did add a bit more to line 64.  What are your thoughts?  
```suggestion
//. The next step is to define a `def` function for the environment using the
//. `create` method:
```

davidchambers: This works for me, but for the fact that I consider `$.create` to be a function rather than a method. How about simply appending _using `$.create`_ to the sentence?

mreinigjr: So where you thinking something like this?

> The next step is to define a `def` function for the environment using `$.create`.

or where you thinking that we put the type signature in there?  So it would look something like this:

> The next step is to define a `def` function for the environment using `$.create`, which has the signature: `create :: { checkTypes :: Boolean, env :: Array Type } -> String -> StrMap (Array TypeClass) -> Array Type -> Function -> Function`.

I suppose we could combine the two options above into this:

> The next step is to define a `def` function for the environment using `$.create :: { checkTypes :: Boolean, env :: Array Type } -> String -> StrMap (Array TypeClass) -> Array Type -> Function -> Function`.

davidchambers: I don't think providing the type signature of `$.create` would add value, given our intention to annotate `def` in the code block that follows.

mreinigjr: understood.  I will adjust and submit shortly.  Thank you.

index.js
```diff
@@ -72,6 +73,7 @@
 //. during development. For example:
 //.
 //. ```javascript
+//. //    def :: String -> StrMap (Array TypeClass) -> Array Type -> Function -> Function
```
davidchambers, line 76: Shall we move this to the preceding code block, or include it in both places?

index.js
```diff
@@ -83,10 +85,10 @@
...
 //. //    add :: Number -> Number -> Number
 //. const add =
-//. def ('add')
-//.     ({})
-//.     ([$.Number, $.Number, $.Number])
-//.     (x => y => x + y);
+//. def ('add') // name of function
+//.     ({}) // type-class constraints
+//.     ([$.Number, $.Number, $.Number]) // function argument and output types
+//.     (x => y => x + y); // curried function implementation
```
davidchambers, line 91: Would terser labels be equally informative, do you think?

```suggestion
//. def ('add')                             // name
//.     ({})                                // type-class constraints
//.     ([$.Number, $.Number, $.Number])    // input and output types
//.     (x => y => x + y);                  // implementation
```

index.js
mreinigjr: I think both is a great idea.

index.js
mreinigjr: I think this really adds the clarity I intended, but more concisely.  I will have these changes pushed shortly.

mreinigjr:
@davidchambers, I just wanted to check in.  I believe all changes you wanted have been submitted.  Please let me know if there is anything further I can do on this PR. Thanks!

davidchambers:
Thanks again for this contribution, @mreinigjr. I combined your commits into a single commit, removed the portions of the commit message that are no longer applicable, and rearranged one sentence. I enjoyed writing the following script to demonstrate how the final commit message is derived from two earlier commit messages:

```bash
#!/usr/bin/env bash

function new {
  git --no-pager show ccf3c12663557a9b1dd57c6950077ae1f376607e --no-patch --format=format:'Author: %an <%ae>%n'
  git --no-pager show ccf3c12663557a9b1dd57c6950077ae1f376607e --no-patch --format=format:'Date:   %ad%n%n'
  git --no-pager show ccf3c12663557a9b1dd57c6950077ae1f376607e --no-patch --format=format:'%s%n%n'
  git --no-pager show ccf3c12663557a9b1dd57c6950077ae1f376607e --no-patch --format=format:'%b' | paste -s -d ' ' -
}

function old {
  git --no-pager show 731804397dd69289a1c243a4a850ba9e0dbb99b3 --no-patch --format=format:'Author: %an <%ae>%n'
  git --no-pager show 731804397dd69289a1c243a4a850ba9e0dbb99b3 --no-patch --format=format:'Date:   %ad%n%n'
  git --no-pager show 5822d6d6c93f5e4eb5cd74245878a9a462cb1533 --no-patch --format=format:'%s%n%n'
  git --no-pager show 731804397dd69289a1c243a4a850ba9e0dbb99b3 --no-patch --format=format:'%b' | paste -s -d ' ' - \
  | sed -E \
"s!"\
"(Type signature)s( for the 'def' function )and the 'create' method are"\
"( not well documented here. Type signatures)( for) both( 'def') and 'create'"\
"( when introduced)( have been added)( as well as comments in example)s"\
"(. Hopefully, this adds clarity.).*"\
"!\1\2is\3\7\4\5\6\8\9!"
}

new | shasum -a 256
old | shasum -a 256
```

mreinigjr:
It was my pleasure @davidchambers and thank you for all the feedback.

PR #294, opened 2020-07-18
types: define Buffer

davidchambers:
This was recently requested by @jceb in sanctuary-js/sanctuary#575.

### DIFF
--- a/index.js
+++ b/index.js
@@ -615,6 +615,18 @@
     ([])
     (typeofEq ('boolean'));
 
+  //# Buffer :: Type
+  //.
+  //. Type comprising every [Buffer][] object.
+  var Buffer_ = NullaryTypeWithUrl
+    ('Buffer')
+    ([])
+    (function(x) {
+       return typeof Buffer !== 'undefined' &&
+              // eslint-disable-next-line no-undef
+              Buffer.isBuffer (x);
+     });
+
   //# Date :: Type
   //.
   //. Type comprising every Date value.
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -615,6 +615,18 @@
...
+  //.
+  //. Type comprising every [Buffer][] object.
+  var Buffer_ = NullaryTypeWithUrl
+    ('Buffer')
+    ([])
+    (function(x) {
+       return typeof Buffer !== 'undefined' &&
+              // eslint-disable-next-line no-undef
+              Buffer.isBuffer (x);
+     });
```
davidchambers, line 628: The following implementation is more elegant but less conducive to maintaining 100% branch coverage:

```javascript
(typeof Buffer === 'undefined' ? K (false) : Buffer.isBuffer)
```

test/index.js
```diff
@@ -1552,6 +1553,26 @@ See https://github.com/sanctuary-js/sanctuary-def/tree/v${version}#Array2 for in
...
+
+    (Buffer => {
+      delete global.Buffer;
+      try {
+        eq (isBuffer (null)) (false);
+        eq (isBuffer (Buffer.from ([1, 2, 3]))) (false);
+      } finally {
+        global.Buffer = Buffer;
+      }
+    }) (Buffer);
```
davidchambers, line 1573: `delete global.Buffer` removes our ability to reference `Buffer` implicitly, so we must retain an explicit reference. `(Buffer => …) (Buffer)` strikes me as more elegant than `const Buffer_ = Buffer; …`.

Avaq: What about `const Buffer = global.Buffer`?

Avaq:
The code and approach look good to me. However, I would've probably separated any environment-specific types out into their own  packages. I could imagine `sanctuary-html`, `sanctuary-node`, and `sanctuary-deno` packages that each provide `env`s and utilities for their corresponding environments. Sanctuary-def's `env` would then only contain JavaScript's built-in types.

davidchambers:
Providing separate packages for various environments is certainly an option worth considering. :)

test/index.js
davidchambers: Good idea. I did not consider `{ const Buffer = global.Buffer; … }`.

PR #295, opened 2020-08-30
preserve field ordering when formatting record types

davidchambers:
The fields of a record type may be specified in some logical order, so it is a shame to reorder the fields in the type's string representation. This pull request removes the use of `Array#sort`; a record type's fields are now enumerated in their inherent order rather than in alphabetical order. This also affects the order in which type checking is performed.

If one is defining a record type whose fields are not logically related, one can of course *specify* the fields in alphabetical order.

If this pull request is merged, equivalent record types may have different string representations despite being equal according to `Z.equals`. Although this is unfortunate, it is not without precedent:

```javascript
> Z.equals (0, -0)
true

> show (0)
'0'

> show (-0)
'-0'
```

`Z.equals` does not even guarantee the interchangeability of equal values with the *same* string representation:

```javascript
> Z.equals ({}, Object.create (null))
true

> 'toString' in {}
true

> 'toString' in Object.create (null)
false
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -411,7 +411,8 @@
         Z.equals (this.name, other.name) &&
         Z.equals (this.url, other.url) &&
         Z.equals (this.supertypes, other.supertypes) &&
-        Z.equals (this.keys, other.keys) &&
+        this.keys.length === other.keys.length &&
+        this.keys.every (function(k) { return other.keys.includes (k); }) &&
         Z.equals (this.types, other.types)
       );
     }
@@ -1917,7 +1918,7 @@
   //. //   The value at position 1 is not a member of ‘{ x :: FiniteNumber, y :: FiniteNumber }’.
   //. ```
   function RecordType(fields) {
-    var keys = sortedKeys (fields);
+    var keys = Object.keys (fields);
 
     function format(outer, inner) {
       if (isEmpty (keys)) return outer ('{}');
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -411,7 +411,8 @@
         Z.equals (this.name, other.name) &&
         Z.equals (this.url, other.url) &&
         Z.equals (this.supertypes, other.supertypes) &&
-        Z.equals (this.keys, other.keys) &&
+        this.keys.length === other.keys.length &&
+        this.keys.every (function(k) { return other.keys.includes (k); }) &&
```
davidchambers, line 415: This assumes that `this.keys` and `other.keys` are set-like arrays. I believe this to be a safe assumption.

test/index.js
```diff
@@ -1102,9 +1106,9 @@ See https://github.com/sanctuary-js/sanctuary-def/tree/v${version}#Number for in
     throws (() => { length ({start: 0, end: 0}); })
            (new TypeError (`Invalid value
 
-length :: { end :: { x :: Number, y :: Number }, start :: { x :: Number, y :: Number } } -> Number
-                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-                                1
+length :: { start :: { x :: Number, y :: Number }, end :: { x :: Number, y :: Number } } -> Number
+                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+                                  1
```
davidchambers, line 1111: This shows the effect of the change on type checking.

Issue #300, opened 2021-06-15
CommonJS environment without "util" module


dotnetCarpenter:
I'm trying to get some help to refactor my code and scrimba.com is awesome at that since you can follow code edits like a video.

But sanctuary-def uses the node.js module `util` and expect all CommonJS environments to be node.js.

I get the following error when I try to run my code at https://scrimba.com/scrim/cQGBN6cM:

> Error: Unknown require: util (in module /node_modules/sanctuary-def/index.js) (/node_modules/sanctuary-def/index.js:191)

Previously I got help from @Avaq here: https://scrimba.com/scrim/cRgK3yAr (click play down in the left corner). In that example the code was not meant to run but I'm trying in the first example, where I am getting the error.

PR #301, opened 2021-06-15
make sure that sanctuary-def will run without "util" module

dotnetCarpenter:
Fix so sanctuary-def can be used in CommonJS enviroments that does
not have the `util` module, such as scrimba.com.

Here is an example with sanctuary-def 0.22.0:
https://scrimba.com/scrim/cQGBN6cM

The def test "returns a function with "inspect" and "@@show" methods"
fails with the default `util` object but I guess you already know
that.

This will hopefully fix issue #300 but I guess the only way to know for sure is to publish and test it out on scrimba.com.

### DIFF
--- a/index.js
+++ b/index.js
@@ -186,9 +186,19 @@
 
   var util = {inspect: {}};
 
+  function alt(a, f) {
+    try {
+      return f ();
+    } catch (e) {
+      /* istanbul ignore next */
+      return a;
+    }
+  }
+
   /* istanbul ignore else */
   if (typeof module === 'object' && typeof module.exports === 'object') {
-    module.exports = f (require ('util'),
+    module.exports = f (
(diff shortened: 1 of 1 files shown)


dotnetCarpenter:
Closing in favour of:

```js
    wrapped.toString = always0 (typeSignature (typeInfo));

    // Node.js
    if (
      typeof process === 'object' &&
      process != null &&
      process.versions != null &&
      process.versions.node != null
    ) wrapped[Symbol.for('nodejs.util.inspect.custom')] = wrapped.toString;

    // Deno
    if (
      typeof Deno !== 'undefined' &&
      Deno != null &&
      typeof Deno.customInspect === 'symbol'
    ) wrapped[Deno.customInspect] = wrapped.toString;
```

PR #303, opened 2021-06-16
provide custom `inspect` behaviour in Node without using `require`

### LINKED ISSUE #300: CommonJS environment without "util" module

davidchambers:
Fixes #300

@dotnetCarpenter and I came up with this solution last night on Gitter. :)

Assuming that CommonJS implies Node.js has caused people problems in several environments. It's exciting to be able to provide nice string representations of `def`-defined functions in the REPL without using the problematic `require ('util')`!

### DIFF
--- a/index.js
+++ b/index.js
@@ -184,12 +184,9 @@
 
   'use strict';
 
-  var util = {inspect: {}};
-
   /* istanbul ignore else */
   if (typeof module === 'object' && typeof module.exports === 'object') {
-    module.exports = f (require ('util'),
-                        require ('sanctuary-either'),
+    module.exports = f (require ('sanctuary-either'),
                         require ('sanctuary-show'),
                         require ('sanctuary-type-classes'),
                         require ('sanctuary-type-identifiers'));
@@ -198,18 +195,15 @@
              'sanctuary-show',
              'sanctuary-type-classes',
              'sanctuary-type-identifiers'],
-            function(Either, show, Z, type) {
-              return f (util, Either, show, Z, type);
-            });
+            f);
   } else {
-    self.sanctuaryDef = f (util,
-                           self.sanctuaryEither,
+    self.sanctuaryDef = f (self.sanctuaryEither,
                            self.sanctuaryShow,
                            self.sanctuaryTypeClasses,
                            self.sanctuaryTypeIdentifiers);
   }
 
-} (function(util, Either, show, Z, type) {
+} (function(Either, show, Z, type) {
 
   'use strict';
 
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -13,6 +13,9 @@
     "release": "sanctuary-release",
     "test": "npm run lint && sanctuary-test && npm run doctest"
   },
+  "engines": {
+    "node": ">=10.12.0"
+  },
```
davidchambers, line 18: <https://nodejs.org/en/blog/release/v10.12.0/>:

>   - __util__
>       - The `util.inspect.custom` symbol is now defined in the global symbol registry as `Symbol.for('nodejs.util.inspect.custom')`. nodejs/node#20857

Avaq (approved):
I followed along with the conversation you had with @dotnetCarpenter. LGTM! :)

index.js
```diff
@@ -2780,7 +2770,14 @@
       } :
       wrapNext ({}, [], 0);
 
-    wrapped[inspect] = wrapped.toString = always0 (typeSignature (typeInfo));
+    wrapped.toString = always0 (typeSignature (typeInfo));
+    /* istanbul ignore else */
+    if (
+      typeof process !== 'undefined' &&
```
dotnetCarpenter, line 2776: @davidchambers This is a great addition to sanctuary-def - I will try it out on scrimba.com soon.

I only want to add that `process !== 'undefined'  && process != null` seems redundant. 

`process != null` should be enough.

```js
null == undefined // true
null != undefined // false
```

[The Abstract Equality Comparison Algorithm](https://262.ecma-international.org/5.1/#sec-11.9.3) says in step 2: _If x is null and y is undefined, return true._ Step 3 is similar but with `x` and `y` switched.

dotnetCarpenter: ah, you don't want to create a global variable called `process` in non-node environments... Got it! Please disregard my comment ;)

dotnetCarpenter: Also I think you pointed out in the chat that not guarding `process` will throw an error....

davidchambers: The expression is verbose, but I don't think any part of it can safely be omitted.

Issue #274, opened 2019-11-24
Wish there was even more detailed error handling in sanctuary-def


voxbono:
I have a use case where I have a RecordType like this:
```Javascript
const newProductRequestBodyType = $.RecordType({
    product_number: $.NonEmpty ($.String),
    name: $.NonEmpty ($.String),
    category: $.NonEmpty ($.String),
    sub_category: $.NonEmpty ($.String),
    supplier: $.NonEmpty ($.String),
    unit: $.NonEmpty ($.String),
    volume_per_unit: $.NonEmpty ($.String)
});
```
And I use this function to validate it:
```Javascript
const isValidObjectByType = recordType => tagBy(is(recordType));
```
`newProductRequestBodyType` is a post request, where I want to validate that each field is valid. Is there a simple way, to provide an error message for each field, so that I can show the right error message to the form on the client?
The `Left` just gives me the sent object back, but it doesn't say why the validation of the Recordtype failed, if that makes sense?

It would be super-useful if the Left contained more details about the error.... Maybe even something I define myself when I create my RecordType.

(This issue was first mentioned on Gitter).

dotnetCarpenter:
This feature is much more popular than I initial thought. #220 and #219 are also about this.

### Part I

@davidchambers posted a proof-of-concept pure function on gitter that provides good error messages in English. The drawbacks being that it's painful to translate the error messages into user facing messages.

As mention in both #219 and #220, there already exist an exposed (though supposedly private) method to validate types on each type. My initial thought was to translate the internal (to sanctuary-def) error message into 3 well-define error messages.

As far as I know there is only 5 possible outcome of an validation of any type (unless we bake in interconnected rules, which we do not).

1. The value is `undefined`/`null`.
2. Properties are missing from the value.
3. Properties are extraneous (which we will ignore as is custom in JS land).
4. Properties are wrong. 
5. A value is correct.

Currently we have `Type ~> validate -> Env -> a`.

Example:
```js
const BarFoo = $.NamedRecordType
  ('BarFoo')
  ('')
  ([])
  ({
    foo: $.String,
    bar: $.NonEmpty (
      $.NullaryType
        ('DateIso')
        ('')
        ([$.String])
        (x => {
          console.debug (`test ran with x = ${x}`);
          return /^\d{4}-[0-1]\d-[0-3]\d$/.test (x);
        })
    ),
    fez: $.Number,
  });

BarFoo.validate ($.env) (barFoo);
```

[...]

PR #308, opened 2021-12-24
validate 🎁🎅

dotnetCarpenter:
### Signature

```haskell
validate :: Type -> a -> Either Object a
```

### Example

```js
// Right (undefined)
const valid   = $.validate ($.Undefined) (undefined);

// Left ({"error": "WrongValue", "name": "$$", "type": "Undefined", "value": 1})
const invalid = $.validate ($.Undefined) (1);
```

### Description

Takes a type, and any value. Returns an `Array` of
`Right a` if the value is a member of the type;
`Left Object` for each property that is invalid.
The first index in an array of `Left`s is always named `$$`,
which refers to the entire value.

### Custom Type Example

```js
//    $DateIso :: NullaryType
const $DateIso = (
  $.NullaryType ('DateIso')
                ('https://www.w3.org/QA/Tips/iso-date')
                ([$.String])
                (x => /^\d{4}-(?:0[1-9]|1[0-2])-(?:0[1-9]|[1-2]\d|3[0-1])$/.test (x))
);

const model = $.RecordType ({
  date: $.NonEmpty ($DateIso),
  bool: $.Boolean,
});

// Right ({"bool": false, "date": "2020-04-10"})
const valid   = $.validate (model) ({date: '2020-04-10', bool: false});

/* [
  Left ({"error": "WrongValue", "name": "$$", "type": "RECORD", "value": {"bool": "foobar", "date": "2020-04-100"}}),
  Left ({"error": "WrongValue", "name": "date", "type": "DateIso", "value": "2020-04-100"}),
  Left ({"error": "WrongValue", "name": "bool", "type": "Boolean", "value": "foobar"}),
] */
const invalid = $.validate (model) ({date: '2020-04-100', bool: 'foobar'});

/* [
  Left ({"error": "WrongValue", "name": "$$", "type": "RECORD", "value": {"bool": false}}),
  Left ({"error": "MissingValue", "name": "date", "type": "NonEmpty", "value": undefined}),
] */
const invalid = $.validate (model) ({bool: false});
```

Ref: https://github.com/sanctuary-js/sanctuary-def/issues/274


index.js
```diff
@@ -1499,6 +1499,69 @@
...
+          var type      = t.types[propName];
+          var result    = type.validate ([]) (propValue);
+
+          if (result.isLeft) {
+            if (propName in x) {
+              returnValue.push (
+                Left ({
+                  error: 'WrongValue',
+                  // TODO: figure what propPath really is
+                  type: result.value.propPath.length > 0
```
dotnetCarpenter, line 1540: @davidchambers this part I'm really not sure about. The idea is to get the name of the Type that is invalid for nested types. I haven't seen any `propPath` that didn't exactly contain `['$1']`but I'm sure there are cases.

index.js
```diff
@@ -1499,6 +1499,69 @@
...
+                    ? type.types[result.value.propPath[0]].name
+                    : type.name,
```
dotnetCarpenter, line 1542: @davidchambers this part I'm really not sure about. The idea is to get the name of the Type that is invalid for nested types. I haven't seen any `propPath` that didn't exactly contain `['$1']` but I'm sure that there are cases.

index.js
```diff
@@ -1499,6 +1506,105 @@
...
+          if (p.name in x) {
+            return Left ({
+              error: 'WrongValue',
+              // TODO: figure out what propPath really is
+              type: p.result.value.propPath.length > 0
+                ? p.type.types[p.result.value.propPath[0]].name
+                : p.type.name,
+              name: p.name,
+              value: p.value
+            });
```
dotnetCarpenter, line 1555: @davidchambers this part I'm really not sure about. The idea is to get the name of the Type that is invalid for nested types. I haven't seen any `propPath` that didn't exactly contain `['$1']` but I'm sure that there are cases.

dotnetCarpenter:
Since a value can have 3 distinct outcomes, I think that it would make the code more readable to encode that as a sum type.

1. `Right a` A value is correct.
2. `Left MissingValue` Property is missing on the value.
3. `Left WrongValue` Value is wrong or property value on the value is wrong.

@davidchambers  How do you feel about requiring [daggy](https://github.com/fantasyland/daggy) in sanctuary-def?
I haven't done anything with daggy yet and won't try if you don't feel it should be a dependency of sanctuary-def.

test/index.js
```diff
@@ -3877,3 +3877,140 @@ suite ('interoperability', () => {
...
+       ]);
+
+    eq ($.validate (model2) ({date: [], bool: false}))
+       ([
+        Left ({'error': 'WrongValue', 'name': '$$', 'type': 'RECORD', 'value': {'bool': false, 'date': []}}),
+        Left ({'error': 'WrongValue', 'name': 'date', 'type': 'NonEmpty', 'value': []}),
+       ]);
+
+    eq ($.validate (model2) ({bool: false, date: '2020-04-10'}))
+       ([Right ({date: '2020-04-10', bool: false})]);
```
davidchambers, line 4012: `Array (Either ValidationError a)` permits invalid combinations such as `[Left (e), Right (x)]` and `[Right (x), Right (x), Right (x)]`. We could make such invalid combinations unrepresentable by changing the nesting of the types:

```haskell
Either (Array ValidationError) a
```

What do you think of this option?

dotnetCarpenter: I'm wondering if there is a use-case for having `Right (x)`s when `$$` fails. Basically removing [step 4](https://github.com/sanctuary-js/sanctuary-def/pull/308/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R1592-R1595).

```js
/* [
Left ({"error": "WrongValue", "name": "$$", "type": "RECORD", "value": {"bool": false}}),
Left ({"error": "MissingValue", "name": "date", "type": "NonEmpty", "value": undefined}),
Right ({"name": "bool", "result": Right (false), "type": Boolean, "value": false})
] */
$.validate (model2) ({bool: false});
```

But if there is not then I think your suggestion is great!

I still have 3 questions though.

1. How do I define `ValidationError`? As a POJO (as in this PR), via `def` or using daggy?
2. is `ValidationError` a sum type of `MissingValue` and `WrongValue`? _See below †_
3. What is the possible values contained in `p.result.value.propPath`? Can it be hard-coded as `p.type.types.$1`?

† I'm not super familiar with sum types but as I understand, the encoding would be something like below, for returning an array of `Either`s:

```reason
type Error   = Left (ValidationError);
type Success = Right (x);

type result  = Error (String, String, String, x)
             | Success (x);

let validate = (Type, x) => {
	/* validation logic */
	[result, result, result, ...result];
};
```

davidchambers: > I'm wondering if there is a use-case for having `Right (x)`s when `$$` fails.

In my experience it's rare that one is interested in what *succeeded* when validation fails.

Let's change the nesting. :)

dotnetCarpenter:
There is plenty of room for optimization but before I do any of that, I want to create a simple form validation example using the new  `$.validate`.

@davidchambers I really need you help to understand [`propPath`](https://github.com/sanctuary-js/sanctuary-def/pull/308#pullrequestreview-840033144).
I might also need your help with [sanctuary-benchmark](https://github.com/sanctuary-js/sanctuary-benchmark). I looked at it before for https://github.com/dotnetCarpenter/express-mime-sniff but was not able to get anything working.

Issue #309, opened 2022-01-04
HtmlElement is unrecognized


dotnetCarpenter:
I believe I have found a bug in how `$.HtmlElement ` is defined.

```html
<section id="HtmlElement"></section>
```

```js
import $ from 'sanctuary-def'

const HtmlElement = document.querySelector ('#HtmlElement')

const testCase = $.test ([])
                        ($.HtmlElement)
                        (HtmlElement)

const sanityCheck = $.test ([])
                           ($.Boolean)
                           (true)

/*
  The Console should display true since a <section> is an HTMLElement.
  All test-cases should be true
 */

console.debug ('testCase', testCase)       // false
console.debug ('sanityCheck', sanityCheck) // true
console.debug ('check', /^\[object HTML.+Element\]$/.test (Object.toString.call (HtmlElement)))
```

Live test-case: https://playcode.io/849895/
Offending line: https://github.com/sanctuary-js/sanctuary-def/blob/master/index.js#L719

This means that currently one can not use an HtmlElement as an argument.

_please ignore the silliness of this example - I realize that `HtmlElement` is not a `StrMap`_

```js
const textContent = S.flip (S.insert ('textContent'))

textContent (HtmlElement) ('text content to inserted')
```

```
Uncaught TypeError: Unrecognized value

flip :: Functor f => f (a -> b) -> a -> f b
                                   ^
                                   1

1)  <object HTMLElement> :: (no types)

The value at position 1 is not a member of any type in the environment.

The environment contains the following types:

  - Function
  - Arguments
  - Array c
  - Array2 c d
  - Boolean
  - Buffer
  - Date
  - Descending c
  - Either c d
  - Error
  - (c -> d)
  - HtmlElement
  - Identity c
  - JsMap c d
  - JsSet c
  - Maybe c
  - Module
  - Null
  - Number
  - Object
  - Pair c d
  - RegExp
  - StrMap c
  - String
  - Symbol
  - Type
  - TypeClass
  - Undefined
  - Future c d
  - ConcurrentFuture c d
  - Event
    invalidValue index.js:2666
    satisfactoryTypes index.js:1351
    assertRight index.js:2731
    wrapNext index.js:2823
```

dotnetCarpenter:
A fix could be to change [line 718-720](https://github.com/sanctuary-js/sanctuary-def/blob/master/index.js#L719) to

```js
function(x) {
[...]

PR #310, opened 2022-01-04
fix test for HtmlElement type

dotnetCarpenter:
This PR addresses the issue of `HtmlElement`'s test not working for HTMLElements as described in #309.

Currently, the test will always fail because `Object.toString.call` on an HTMLElement produces an error.

In Gecko (Firefox) the error is: "TypeError: Function.prototype.toString called on incompatible object"

In Blink (Chrome, Edge) the error is: "TypeError: Function.prototype.toString requires that 'this' be a Function"

This PR does not contain any new unit tests.

### DIFF
--- a/index.js
+++ b/index.js
@@ -716,7 +716,11 @@
     ('HtmlElement')
     ([])
     (function(x) {
-       return /^\[object HTML.+Element\]$/.test (toString.call (x));
+      return (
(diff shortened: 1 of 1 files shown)


dotnetCarpenter:
Obsoleted by #311 

I used `Object.toString;` and **not** `Object.prototype.toString;`.

PR #311, opened 2022-01-05
types: fix HtmlElement predicate

### LINKED ISSUE #309: HtmlElement is unrecognized

davidchambers:
Fixes #309

```javascript
> Object.prototype.toString.call (document.createElement ('div'))
'[object HTMLDivElement]'

> Object.prototype.toString.call (document.createElement ('section'))
'[object HTMLElement]'
```

Before:

```javascript
> /^\[object HTML.+Element\]$/.test (Object.prototype.toString.call (document.createElement ('section')))
false
```

After:

```javascript
> /^\[object HTML.*Element\]$/.test (Object.prototype.toString.call (document.createElement ('section')))
true
```

/cc @dotnetCarpenter

### DIFF
--- a/index.js
+++ b/index.js
@@ -716,7 +716,7 @@
     ('HtmlElement')
     ([])
     (function(x) {
-       return /^\[object HTML.+Element\]$/.test (toString.call (x));
+       return /^\[object HTML.*Element\]$/.test (toString.call (x));
      });
 
   //# Identity :: Type -> Type
--- a/test/index.js
+++ b/test/index.js
@@ -1672,6 +1672,10 @@ See https://github.com/sanctuary-js/sanctuary-def/tree/v${version}#Array2 for in
     eq ($.HtmlElement.name) ('HtmlElement');
     eq ($.HtmlElement.url) (`https://github.com/sanctuary-js/sanctuary-def/tree/v${version}#HtmlElement`);
     eq ($.HtmlElement.supertypes) ([]);
+
+    const isHtmlElement = $.test ([]) ($.HtmlElement);
+    eq (isHtmlElement ({[Symbol.toStringTag]: 'HTMLDivElement'})) (true);
+    eq (isHtmlElement ({[Symbol.toStringTag]: 'HTMLElement'})) (true);
   });
 
   test ('provides the "Identity" type constructor', () => {


dotnetCarpenter:
@davidchambers 👍 

I've tested your change in Firefox and it works! But I am puzzled to how I got ... aha I see. I used `Object.toString;` and **not** `Object.prototype.toString;`.

You're fix is great :)

index.js
```diff
@@ -716,7 +716,7 @@
     ('HtmlElement')
     ([])
     (function(x) {
-       return /^\[object HTML.+Element\]$/.test (toString.call (x));
+       return /^\[object HTML.*Element\]$/.test (toString.call (x));
```
dotnetCarpenter, line 719: How about disallow space here? `^\[object HTML[^\s]*Element\]$`

Avaq (approved):

index.js
dotnetCarpenter: @Avaq What do you think about changing the regex to `^\[object HTML[^\s]*Element\]$` in order to disallow spaces after `HTML`?

So `HTML ElementDiv` would fail while `HTML.ElementDiv` would be allowed.

davidchambers: We will accept infinitely many invalid strings in either case; I'm inclined to stick with the simpler `.*`.

By the way, `\S` is shorthand for `[^\s]`. :)

Avaq: > We will accept infinitely many invalid strings in either case

That was my thinking too.

dotnetCarpenter: I guess, since `HTMLElement` is an exotic object, baking in too many assumptions is bound to fail in edge cases.

The real test-case would be `return x instanceof HTMLElement` of course, but I see how that is difficult to test in an environment without `HTMLElement`, like nodejs.

Sanctuary-def's test of `HTMLElement` is good enough so that Sanctuary works client-side without requiring users to create a custom `HTMLElement` definition.

PR #313, opened 2023-06-30
types: define BigInt

eush77:
This PR defines `BigInt` type.

Since it requires ES2020, update ESLint configuration for testing. This triggered `comma-dangle` in a bunch of places for me, so I fixed those, too.

The actual functional change in `index.js` does not require ES2020.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,6 +1,8 @@
 {
   "root": true,
   "extends": ["./node_modules/sanctuary-style/eslint-es6.json"],
+  "env": {"es2020": true},
+  "parserOptions": {"ecmaVersion": 11},
   "overrides": [
     {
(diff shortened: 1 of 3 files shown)


davidchambers:
Thank you very much for this contribution, @eush77. I'm sorry for taking so long to review it!

I would like to reduce the scope of the changes. If we remove `1n` from the test suite, setting `env.es2020` to `true` should suffice. We need not test the literal form given that `BigInt (1)` evaluates to the same value.

What do you think? I realize that months have passed and you've no doubt moved on to other things. If you'd like to update the pull request yourself, I promise not to leave you hanging again. If you'd prefer for me to make the changes, just say so.

Thanks again for making Sanctuary better. :heart:

PR #323, opened 2023-12-30
use `globalThis` to obviate the need for trailing underscores

davidchambers:
By using `globalThis.Array` we are able to write `const Array = ...;` then reference `Array` throughout the file. Ditto for the names of other built-in types.

### DIFF
--- a/index.js
+++ b/index.js
@@ -203,13 +203,13 @@
                            self.sanctuaryTypeIdentifiers);
   }
 
-}) ((Either, show, Z, type) => {
+}) ((E, show, Z, type) => {
 
   'use strict';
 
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -520,8 +520,8 @@
   //. Type comprising every [`Date`][] value except `new Date (NaN)`.
   const ValidDate = NullaryTypeWithUrl
     ('ValidDate')
-    ([Date_])
-    (B (complement (Number.isNaN)) (Number));
+    ([Date])
+    (date => !(globalThis.Number.isNaN (date.valueOf ())));
```
davidchambers, line 524: I chose to rewrite this predicate without `B` and `complement`.

Avaq (approved):

PR #325, opened 2023-12-30
use ES modules

davidchambers:
sanctuary-js/sanctuary#726


test/index.js
```diff
@@ -81,6 +82,8 @@ def ('$26')
     (a => b => c => d => e => f => g => h => i => j => k => l => m => n => o => p => q => r => s => t => u => v => w => x => y => z =>
        [a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p, q, r, s, t, u, v, w, x, y, z]);
 
+const suite = (_, run) => run ();
```
davidchambers, line 85: This is a small hack to keep the tests running with minimal changes. I have noticed that this indirection makes the locations of test failures reported by Oletus less helpful; I am willing to tolerate this.

test/module.js
```diff
@@ -11,4 +9,4 @@ $.create
   ({})
   ([a, a])
   (x => x)
-  (fs);
+  ($);
```
davidchambers, line 12: `$` is now a `Module` object so can be used in place of `fs`.

index.js
```diff
@@ -180,2669 +180,2588 @@
...
+      ([Type, Type])
+      (_UnaryType (name) (functionUrl (name)) (supertypes) (test) (_1))
+);
+
+const BinaryTypeWithUrl = name => supertypes => test => _1 => _2 => (
+  def (name)
+      ({})
+      ([Type, Type, Type])
+      (_BinaryType (name) (functionUrl (name)) (supertypes) (test) (_1) (_2))
+);
```
davidchambers, line 370: These functions now do more than before: they use `def` in addition to `functionUrl`.

index.js
```diff
@@ -180,2669 +180,2588 @@
...
-    INCONSISTENT,
-    '',
-    '',
-    0,
-    (outer, inner) => '???',
-    [],
-    null,
-    []
-  );
+const env = [];
```
davidchambers, line 384: `env` is now defined here, provided to `mkdef` below, and populated later. :see_no_evil:

index.js
```diff
@@ -180,2669 +180,2588 @@
...
-    env => x => true,
-    []
-  );
+//# Type :: Type
+//.
+//. Type comprising every `Type` value.
+export const Type = NullaryTypeWithUrl
+  ('Type')
+  ([])
+  (x => type (x) === 'sanctuary-def/Type@1');
```
davidchambers, line 400: I moved this block to the top of the section as other type definitions depend on `Type`.

index.js
```diff
@@ -180,2669 +180,2588 @@
...
+//. the type comprising every [`String`][] value except `''`.
+//.
+//. The given type must satisfy the [Monoid][] and [Setoid][] specifications.
+export const NonEmpty = UnaryTypeWithUrl
+  ('NonEmpty')
+  ([])
+  (x => Z.Monoid.test (x) &&
+        Z.Setoid.test (x) &&
+        !(Z.equals (x, Z.empty (x.constructor))))
+  (monoid => [monoid]);
```
davidchambers, line 414: I moved this block near the top of the section as other type definitions depend on `NonEmpty`.

index.js
```diff
@@ -180,2669 +180,2588 @@
...
-  const fromUncheckedUnaryType = typeConstructor => {
-    const t = typeConstructor (Unknown);
-    return def (t.name) ({}) ([Type, Type]) (fromUnaryType (t));
-  };
-
-  //  fromUncheckedBinaryType :: (Type -> Type -> Type) -> Type -> Type -> Type
-  const fromUncheckedBinaryType = typeConstructor => {
-    const t = typeConstructor (Unknown) (Unknown);
-    return def (t.name) ({}) ([Type, Type, Type]) (fromBinaryType (t));
-  };
```
davidchambers, line 2689: These functions are no longer used.

index.js
```diff
@@ -180,2669 +180,2588 @@
...
+                  }
+                  return ' '.repeat (s.length);
+                })
+  );
+};
+
+//  satisfactoryTypes :: ... -> Either (() -> Error)
+//                                     { typeVarMap :: TypeVarMap
+//                                     , types :: Array Type }
+function satisfactoryTypes(
```
davidchambers, line 1225: This function now needs to be hoisted. :see_no_evil:

index.js
```diff
@@ -180,2669 +180,2588 @@
...
+  Pair (Unknown) (Unknown),
+  RegExp,
+  StrMap (Unknown),
+  String,
+  Symbol,
+  Type,
+  TypeClass,
+  Undefined,
+);
+export {env};
```
Avaq, line 1000: Why not have the `export` at the `const` statement?

Avaq (approved):

test/index.js
Avaq: See also https://github.com/bearror/oletus/issues/21

index.js
davidchambers: I tried to move as few things as possible from their current—presumably intended—locations. I moved `env` because I could see no other option.

Whether we export `env` here or there is not important to me. Do you have a preference?

Avaq: My preference would be to have the export of `env` by its definition, irrespective of whether it's being mutated later.

PR #326, opened 2023-12-30
broaden and future-proof `RegexFlags`

davidchambers:
`RegexFlags` is woefully out of date. This pull request updates the type's predicate and obviates the need for future updates.

### DIFF
--- a/index.js
+++ b/index.js
@@ -896,16 +896,15 @@ export const String = NullaryTypeWithUrl
 
 //# RegexFlags :: Type
 //.
-//. Type comprising the canonical RegExp flags *recognized by the runtime*.
-//. Characters must be alphabetically ordered. Repeated characters are not
-//. permitted. Invalid combinations (such as `'uv'`) are not permitted.
+//. Type comprising the RegExp flags *accepted by the runtime*.
+//. Repeated characters are not permitted. Invalid combinations
+//. (such as `'uv'`) are not permitted.
 export const RegexFlags = NullaryTypeWithUrl
   ('RegexFlags')
   ([String])
   (s => {
-     let regex;
-     try { regex = globalThis.RegExp ('', s); } catch { return false; }
-     return regex.flags === s;
+     try { globalThis.RegExp ('', s); } catch { return false; }
+     return true;
    });
 
 //# Symbol :: Type
--- a/index.js
+++ b/index.js
@@ -896,20 +896,17 @@ export const String = NullaryTypeWithUrl
 
 //# RegexFlags :: Type
 //.
-//. Type comprising the canonical RegExp flags:
-//.
-//.   - `''`
-//.   - `'g'`
-//.   - `'i'`
-//.   - `'m'`
-//.   - `'gi'`
-//.   - `'gm'`
-//.   - `'im'`
-//.   - `'gim'`
+//. Type comprising the canonical RegExp flags *recognized by the runtime*.
+//. Characters must be alphabetically ordered. Repeated characters are not
+//. permitted. Invalid combinations (such as `'uv'`) are not permitted.
 export const RegexFlags = NullaryTypeWithUrl
   ('RegexFlags')
   ([String])
-  (s => /^g?i?m?$/.test (s));
+  (s => {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -896,20 +896,17 @@ export const String = NullaryTypeWithUrl
...
+//. Characters must be alphabetically ordered. Repeated characters are not
+//. permitted. Invalid combinations (such as `'uv'`) are not permitted.
 export const RegexFlags = NullaryTypeWithUrl
   ('RegexFlags')
   ([String])
-  (s => /^g?i?m?$/.test (s));
+  (s => {
+     let regex;
+     try { regex = globalThis.RegExp ('', s); } catch { return false; }
+     return regex.flags === s;
```
Avaq, line 908: Could the runtime reorder the flags, potentially causing this assertion to fail even though the flags all got accepted?

davidchambers, line 908: [22.2.6.4: get RegExp.prototype.flags][1] specifies the order of characters, so this check ensures that the characters of `s` are in alphabetical order. :)


[1]: https://tc39.es/ecma262/multipage/text-processing.html#sec-get-regexp.prototype.flags

test/index.js
```diff
@@ -2160,12 +2161,16 @@ See https://github.com/sanctuary-js/sanctuary-def/tree/v${version}#ValidDate for
...
     //  String objects are not acceptable.
     eq (isRegexFlags (new String ('')), false);
     //  Flags must be alphabetically ordered.
     eq (isRegexFlags ('mg'), false);
-    //  "Sticky" flag is not acceptable.
-    eq (isRegexFlags ('y'), false);
+    //  Flags cannot contain repeated characters.
+    eq (isRegexFlags ('gg'), false);
+    //  Flags cannot contain both 'u' and 'v'.
+    eq (isRegexFlags ('uv'), false);
```
Avaq, line 2173: ...it might be worth adding a test for that, because these tests all pass the flags in the same order

index.js
Avaq: Oh, haha, we commented about this at the same time

test/index.js
```diff
@@ -2160,12 +2161,16 @@ See https://github.com/sanctuary-js/sanctuary-def/tree/v${version}#ValidDate for
     eq (isRegexFlags ('gm'), true);
     eq (isRegexFlags ('im'), true);
     eq (isRegexFlags ('gim'), true);
+    eq (isRegexFlags ('dgimsuy'), true);
+    eq (isRegexFlags ('dgimsvy'), parseInt (process.versions.node, 10) >= 20);
```
davidchambers, line 2165: This shows that `'dgimsvy'` is a member of `RegexFlags` in Node.js v20.0.0 but not in earlier versions.

test/index.js
davidchambers: The existing `'mg'` test case should suffice. :)

index.js
Avaq: My concern still stands, though: Should that check be there? Are users forced to supply the flags in the order specified?

test/index.js
Avaq: Oh, indeed!

index.js
Avaq:
```js
> new RegExp('', 'mg').flags
'gm'
```

Like, why don't we consider `mg` valid, even though the runtime accepts it?

Avaq: See other thread: https://github.com/sanctuary-js/sanctuary-def/pull/326#discussion_r1438692907

davidchambers: > Are users forced to supply the flags in the order specified?

Yes. This preserves the current behaviour (the predicate currently uses `/^g?i?m?$/`). Would you like to remove this requirement?

Avaq: Shouldn't we instead just check that `regex.flags.length === s.length`?

Avaq: For the record: I am in favour of removing that requirement. I don't think it's Sanctuary's job to enforce the order of RegEx flags, and should be left to something like a linter.

Avaq (approved):

